Issue #13, opened 2017-01-09
Allow short and long tests


Gbury:
When writing automated tests, is is recommended to have two version of a testsuite: a short one that runs reasonably fast (so that it is effectively run each time a projet is built), and a long one that might be more exhaustive (but whose running time makes it impossible to run at each build).

To that end, it would be useful to have a mechanism to allow that for qcheck tests (particularly for the functions that generate ounit tests). From the qcheck user point of view, maybe something like `~long_count`  argument to provide with `~count` when creating tests ?

c-cube:
This needs some discussions about design.

- Would the "long" switch apply to the whole testsuite, or only to some tests (in `QCheck_runner`)?
  I suppose it makes more sense, e.g. to run all tests more thoroughly after a dangerous change or
  before a release
- default version of `long_count` would be `count`? or `count*2`, something like that?
- how does this interact with `max_gen`? Do we want `long_count` to instead be a multiplicative factor so that `max_gen` can be scaled, too?

Gbury:
- I'd think the long switch would apply to the whole testsuite. Actually, I'm using the ounit wrappers for tests, and then running the ounit main, so I always run the whole testsuite.
- I'd think that the default for `~long_count` should be `~count`. My reasoning is the following: either the test is very quick, and a relatively high `~count` can already be used (so that the short tests is relatively exhaustive), in which case, I'd think the long test should have the same number of instances, or it is not, and in that case, `~count` will be manually set to a value small enough to allow fast testing, and `~long_count` will have to be set to the desired value for exhaustiveness.
- The interaction with `max_gen`is indeed not trivial. Ideally, most options like it should be duplicated to have a `_long` version, but then it would be hell to specify everything. In that case, a multiplicative factor would make the most sense. The option would then be better name `~factor`or `~long_factor`, and would have a default value of `1`.

PR #14, opened 2017-01-09
Add Long tests

Gbury:
Cf Issue #13

### DIFF
--- a/src/QCheck.ml
+++ b/src/QCheck.ml
@@ -841,7 +841,7 @@ module Test = struct
 
   (* main checking function *)
   let check_cell ?(long=false) ?(call=callback_nil_) ?(rand=Random.State.make [| 0 |]) cell =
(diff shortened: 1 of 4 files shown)


src/QCheck.ml
```diff
@@ -838,13 +840,14 @@ module Test = struct
     | Some n -> n
 
   (* main checking function *)
-  let check_cell ?(call=callback_nil_) ?(rand=Random.State.make [| 0 |]) cell =
+  let check_cell ?(long=false) ?(call=callback_nil_) ?(rand=Random.State.make [| 0 |]) cell =
+    let factor = if long then 1 else cell.long_factor in
```
c-cube: looks like it should be the opposite. 1 if false, long_factor otherwise.

c-cube:
would be nice to have a word on this in the README :-)

PR #16, opened 2017-01-09
Add a step function

Gbury:
This PR adds a callback function called after each run of a test on an instance.

### DIFF
--- a/src/QCheck.ml
+++ b/src/QCheck.ml
@@ -729,9 +729,21 @@ module Test = struct
 
   module R = TestResult
 
+  (* Result of an instance run *)
+  type res =
+    | Success
+    | Failure
+    | Error of exn
+
+  (* Step function, called after each instance test *)
+  type 'a step = string -> 'a cell -> 'a -> res -> unit
+
+  let step_nil_ _ _ _ _ = ()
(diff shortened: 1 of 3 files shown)


Gbury:
Arg... I forgot to remove some other changes I made. Let me try and fix that.

Gbury:
Fixed. You can review, ^^

src/QCheck.ml
```diff
@@ -783,11 +795,17 @@ module Test = struct
     | CR_continue
     | CR_yield of 'a TestResult.t
 
+  (* Get the name of a test *)
+  let name_ cell = match cell.name with
```
c-cube: Not your fault, but maybe this is the right opportunity to have `cell.name: string` and generate fresh names at cell construction (so that at least, anonymous tests are named "<test1>", "<test2>", …).

Gbury: Fresh names like "anon_test_%d" ?

Gbury: Done in commit https://github.com/c-cube/qcheck/pull/16/commits/5f5714479bc2593a1ca93d9dd1b630b8d22a8380

PR #19, opened 2017-01-11
Improved reporting for running tests

Gbury:
This Pull requests improves the reporting when running tests using QCheck_runner (for the moment, it doesn't affect the part about ounit{,2}). It also adds an output example in the documentation of the `QCheck_runner.run_tests_main` function.

Currently, the `verbose` option controls both the printing of:
- each test result (printing the number of generated instances, passed tests, failures and errors)
- each test instance (i.e the line concerning a test's results is updated after each run of na instance)

If need be, there could be a new option controlling the printing of each test instance result, though I haven't yet found a name for it. More importantly, the printing of test instance results, doesn't seem very compatible with the ounit runner `QCheck_runner.run`, which share the parsing of command line options.


c-cube:
Good, thanks a lot!

PR #20, opened 2017-01-11
get_count can return count for long version test

Gbury:


Gbury:
Closing in favor of #22

PR #22, opened 2017-01-12
Add QCheck.get_long_factor

Gbury:


c-cube:
Can you add `@since NEXT_RELEASE` on the new function?

Gbury:
Done, I also added it on `get_count` since it also was a new function.

PR #24, opened 2017-01-26
Print module documentation

jmid:
Added a little documentation for the `Print` module


c-cube:
These annotations are a bit redundant, but they won't harm. Thanks!

PR #29, opened 2017-01-27
Change asserts to raise Invalid_arg (following the doc), and update doc

jmid:
Cleanup #28 to still raise `Invalid_arg` on illegal arguments (as the API doc promises).
Also update the .mli file to reflect the loosened restrictions.

### DIFF
--- a/src/QCheck.ml
+++ b/src/QCheck.ml
@@ -140,13 +140,17 @@ module Gen = struct
 
   let int st = if RS.bool st then - (pint st) - 1 else pint st
   let int_bound n =
-    assert (n >= 0);
-    fun st ->
-      let r = pint st in
-      r mod (n+1)
+    if n < 0
+    then invalid_arg "Gen.int_bound"
+    else
+      fun st ->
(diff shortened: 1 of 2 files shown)


src/QCheck.ml
```diff
@@ -140,13 +140,17 @@ module Gen = struct
 
   let int st = if RS.bool st then - (pint st) - 1 else pint st
   let int_bound n =
-    assert (n >= 0);
-    fun st ->
-      let r = pint st in
-      r mod (n+1)
+    if n < 0
+    then invalid_arg "Gen.int_bound"
```
c-cube: If you don't mind, I'd rather have this style:

```ocaml
if n<0 then invalid_arg "…";
fun st -> …
```

because I'm wary of the dangling else. Same goes for the next one.

jmid: Be careful not to be drawn to the dark side :-D
Ok, adjusted in the follow-up.

c-cube:
Thanks!

Issue #33, opened 2017-05-24
Progress feedback to user


jmid:
Currently, an interactive counter in the UI counts up the number of generated inputs. This works beautifully for succeeding tests and fast shrinkers.
For occasional slow tests (complex properties and exponential(?) shrinkers) the UI provides less feedback. For example, for the `check-fun` tests when they hang on `fold_left fold_right` I just get this:
```
generated; error;  fail; pass / total -     time -- test name
[✗] (   1)    0 ;    1 ;    0 / 1000 --     0.0s -- false bool
[✓] (1000)    0 ;    0 ; 1000 / 1000 --     0.0s -- thrice bool
[✗] (   1)    0 ;    1 ;    0 / 1000 --     0.0s -- map filter
```
Because of the order of the tests I know that `fold_left fold_right` is next, and hence that it must be causing it, but from the UI alone I do not know
 - that this is the case
 - that it found a counterexample immediately and now is taking a long time shrinking.

Perhaps the first can simply be addressed by a suitable `flush` of the output channel.
For the second just printing `Found counterexample, shrinking` (or something like) would be nice to get a feel for the framework's progress. A fancier version could even display some progress measure (or bar) of the shrinking phase.

c-cube:
This makes sense indeed, although it's on a per-runner basis. I'll probably need to modify a bit the callbacks because the test events would become more fine grained.

c-cube:
This should not block 0.6, actually.

c-cube:
I think @Gbury might want to help on this :-)

Gbury:
Sure, i'll have a look as soon as possible. But from what I remember, it would probably require to add some hooks to the existing test functions.

c-cube:
Don't worry, 0.6 is not even merged in opam yet, it's not urgent :-)
More hooks are needed indeed, but it should be possible to maintain compatibility, I think.

Gbury:
I proposed an upgrade for progress feedback in PR #35. @jmid , what's your impression about it, as a user ?

jmid:
Sorry for the delay. I had some other hard deadlines that have stressed me. I just downloaded the upgrade and it is very nice - Thanks!
Question: as a user, what is the meaning of the number printed during shrinking? In principle it could either count all shrinking attempts (the number of times we retest a property after the first failure), or it could count only the succesful attempts (the number of times we retest a property and it still fails).

jmid:
[...]

PR #35, opened 2017-05-31
Better progress feedback in verbose mode

Gbury:
Following comments in issue #33, this pull request make the verbose mode a bit better. The modifications made are:

- Add colors to the checkmarks printed to the left
- Print the status line of a test before it begin (compared to previsouly where it would only be printed after the first instance had been tested)
- Add a notion of event to `QCheck.Test.check_cell`, to represent intermediary events during the testing of an instance
- Use the event handler to print the current status of a test (either generating a an instance, collecting statistics about that instance, testing the law on the instance, or shrinking an instance)

For people interested in seeing the new features in action, I found that using the seed `80394201` makes the `fold_left fold_right uncurried` test take about 3s on my laptop, which is enough to see the progress. On that same example, I didn't notice any significant slowdown due to the event handler.

The current state of the test instance (generating, collecting, testing or shrinking), is currently printed to the right of the test name in parentheses, because I didn't find any other satisfying solution, if anyone has a better idea, I'm open to suggestions.

### DIFF
--- a/src/QCheck.ml
+++ b/src/QCheck.ml
@@ -1142,7 +1142,7 @@ module Test = struct
     | Generating
     | Collecting of 'a
     | Testing of 'a
-    | Shrinked of int * 'a
+    | Shrunk of int * 'a
     | Shrinking of int * int * 'a
 
   type 'a handler = string -> 'a cell -> 'a event -> unit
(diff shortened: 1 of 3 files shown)


c-cube:
I will merge this after the comments on more compact tables is done, if you're ok with this :-)

PR #44, opened 2017-07-06
Update README

jmid:
Here's an update of the README documentation:
- fixed the first example which used `int` while the text actually said `small_nat`
- fixed the tree shrinker which actually only shrunk leaves. I added a minimal printer while doing so.

Generally I supplemented the documentation with some failing tests, which is where one can see the real benefit of printing and shrinking. Also, I tried to elaborate some more about what was going on in the examples.


c-cube:
Thanks! :-)

Issue #41, opened 2017-07-06
Complexity of `add_stat`


jmid:
This can be considered a separate issue to #40 hence the separate report (but feel free to merge).

If I try to inspect the (unsigned) `int` generator (by wrapping it with `abs`) it takes a very long time 
(I still haven't waited it out) to compute statistics of a sample of (the default) 100 integers:
```
# let t = Test.make (add_stat ("dist",abs) int) (fun _ -> true);;
val t : QCheck.Test.t = QCheck.Test.Test <abstr>
# QCheck_runner.run_tests ~verbose:true [t];;                    
generated error  fail pass / total  time(s) name
[✓]  100    0     0    100 /  100      0.0  anon_test_12                                            

(top-level hangs)
```
This behaviour indicates that the 100 `true` tests completed immediately, while computing/drawing statistics take substantial time. I suspect if one of these routines algorithmically "count up" from 0, then it will take a long time to reach the entries produced by `int` (wrapped in `abs`) - but that is just a guess without having studied the code thoroughly.

c-cube:
Indeed, there is a loop on every index. It should instead become a loop on the table `value -> num of occurrences` (which can become quite sparse if values range over a wide domain). This requires to pre-allocate the buckets though, so it will take a bit of time, I can't do it right no.

c-cube:
Fixed by #45, as tests can show. Thanks @jmid!

Issue #40, opened 2017-07-06
Support stats with negative integers


jmid:
Consider the following interaction to observe the distribution of the builtin integer generators:
```
# let t = Test.make (add_stat ("dist", fun x -> x) small_nat) (fun _ -> true);;
val t : QCheck.Test.t = QCheck.Test.Test <abstr>
generated error  fail pass / total  time(s) name
[✓]  100    0     0    100 /  100      0.0  anon_test_8                                             
stats dist:
  num: 100, avg: 332.98, stddev: 1220.61, median 9, min 0, max 8845
            0-442: #######################################################          89
          443-885: ###                                                               5
         886-1328: #                                                                 2
        1329-1771:                                                                   0
        1772-2214:                                                                   0
        2215-2657:                                                                   0
        2658-3100:                                                                   0
        3101-3543:                                                                   1
        3544-3986:                                                                   0
        3987-4429:                                                                   0
        4430-4872:                                                                   0
        4873-5315:                                                                   1
        5316-5758:                                                                   0
        5759-6201:                                                                   0
        6202-6644:                                                                   1
        6645-7087:                                                                   0
        7088-7530:                                                                   0
        7531-7973:                                                                   0
        7974-8416:                                                                   0
        8417-8859:                                                                   1
        8860-9302:                                                                   0
================================================================================
success (ran 1 tests)
- : int = 0
```
Now, it would be nice to similarly inspect a sample of `small_signed_int` (or other signed generators for that matter). However if one attempts to do so you get:
```
# let t = Test.make (add_stat ("dist",fun x -> x) small_signed_int) (fun _ -> true);;
val t : QCheck.Test.t = QCheck.Test.Test <abstr>
# QCheck_runner.run_tests ~verbose:true [t];;                           
generated error  fail pass / total  time(s) name
[ ]    0    0     0      0 /  100      0.0  anon_test_10 (collecting)Exception: Assert_failure ("src/QCheck.ml", 1188, 9).
```
(where the linenumber naturally differs with the particular QCheck version), which indicates that statistics are limited to non-negative integers. In the above case, I would love to see a nice bell-curve centered around 0 as output :-)

This will require some more program logic (for non-negative stats only split into "positive buckets", while mixed positive and negative stats should split into "mixed buckets", ...).

c-cube:
I have a patch that outputs the following disposition. It doesn't solve the performance issue though…

```
Stat for test stats_neg:

stats dist:
[...]

PR #45, opened 2017-07-21
fix complexity of `add_stat`

jmid:
This patch fixes issue #41 the complexity of `add_stat`, so that statistics only take time proportional to the sample size.

In addition 
- it fixes the printing of a single label (prints array index `i` rather than `i'`)
- it fixes a problem with integer overflow
- it reverts the rounding of the first bucket to a multiple of 10. I encountered an issue where 
```
let min_idx =
  if min_idx>0 then 10 * min_idx / 10
  else 10 * (min_idx/10 - (min_idx mod 10))
```
yielded 40 for a `min_idx` of `-4`.

I've run a number of tests(!) for cornercases along the lines of
```
let t = Test.make ~count:1000 (add_stat ("dist",fun x -> x) small_signed_int) (fun _ -> true);;
let t = Test.make ~count:1000 (add_stat ("dist",fun x -> x) small_nat) (fun _ -> true);;
let t = Test.make ~count:1000 (add_stat ("dist",fun x -> x) (int_range (-43643) 435434)) (fun _ -> true);;
let t = Test.make ~count:1000 (add_stat ("dist",fun x -> x) (int_range (-40000) 40000)) (fun _ -> true);;
let t = Test.make ~count:1000 (add_stat ("dist",fun x -> x) (int_range (-4) 4)) (fun _ -> true);;
let t = Test.make ~count:1000 (add_stat ("dist",fun x -> x) (int_range (-4) 17)) (fun _ -> true);;
let t = Test.make ~count:100000 (add_stat ("dist",fun x -> x) int) (fun _ -> true);;
```
each of which can be run with `QCheck_runner.run_tests ~verbose:true [t];;`

The new median computation is O(n log n) (or whatever `List.sort` is) when it could potentially be made O(n) with a suitable [selection algorithm](https://en.wikipedia.org/wiki/Selection_algorithm). For now I consider it a premature optimization to spend too much effort unless it becomes an issue. The last test with a count of a million completes within a second or so from the top-level with the current patch.

I have a separate patch for an aesthetic bucket division (mentioned in #40) which produces a consistent output on consecutive runs. It also prints less empty buckets in the next-to-last example above.

### DIFF
--- a/src/QCheck.ml
+++ b/src/QCheck.ml
@@ -1412,11 +1412,6 @@ module Test = struct
            min i m1, max i m2)
         tbl (max_int,min_int)
     in
-    (* round min_idx to the closest multiple of 10 that is smaller *)
-    let min_idx =
-      if min_idx>0 then 10 * min_idx / 10
-      else 10 * (min_idx/10 - (min_idx mod 10))
-    in
     (* compute average *)
     if !num > 0 then (
       avg := !avg /. float_of_int !num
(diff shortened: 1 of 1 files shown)


c-cube:
This looks nice! Can you please add the tests to the ounit runner test file instead?

c-cube:
I'll do it. Thanks :-)

Issue #55, opened 2018-09-26
Fail when no assumes pass


porglezomp:
I was stubbing out functions, and was surprised when one of my properties started passing. I've figured out that it's simply failing the `assume` every time, and since it never fails, that counts as succeeding.

If not enough assumptions pass, that should be a test failure (or at the very least a warning), since it indicates that your property probably isn't testing the thing you think it's testing very well.

Gbury:
I agree that warning (or raising an error) if no input passed the preconditions seems like a good idea.

That said, if really none of your generated input satisfies the `assume` preconditions, then I'd say that's probably a bug and you should either adjust the `max_gen` parameter of the test, or adapt the generating function to generate less invalid inputs...

c-cube:
Interesting point indeed. I suppose we need at least a warning. Question:

- do we want to have a minimum *number* of successes, or a percentage?

porglezomp:
Before I made the issue, I was checking around for how some other libraries handle this to try to gauge the techniques, and there seem to be a few approaches.

The original Haskell QuickCheck seems to have this be user-directed, and also use a statistical test to estimate the failure probability. FsCheck (at least by default) requires that 100 tests pass assumes, and it stops early at a global limit.

PR #56, opened 2018-09-27
warn/fail if too many preconditions failed

c-cube:
see #55; cc @porglezomp @Gbury 

to test: `make example-runner`, see messages at the beginning.

### DIFF
--- a/example/QCheck_runner_test.ml
+++ b/example/QCheck_runner_test.ml
@@ -53,6 +53,22 @@ let fun2 =
     (fun (QCheck.Fun (_,p)) ->
        not (p "some random string") || p "some other string")
 
+let bad_assume_warn =
+  QCheck.Test.make ~count:2_000
+    ~name:"WARN_unlikely_precond"
+    QCheck.int
+    (fun x ->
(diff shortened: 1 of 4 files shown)


Gbury:
The part about detection looks good, but the printing of errors is problematic.

Specifically, the errors printed heavily mess with the verbose output in the Qcheck_runner_test example. Additionally, these errors are printed out of order (all the other failures are printed in the order of tests, whereas these errors appear at the start of the output in non verbose mode).
 Instead of printing the error on stderr, wouldn't it be simpler to return a failure result (and not a success) for the test ?

c-cube:
Not a precise idea, but probably some opaque type (a record) with a list of warnings inside and a `view` function that returns the normal sum type. Also I'd like to make the sum types use inner records so they're more future proof and readable.

c-cube:
extended the state.

Gbury:
Seems good overall. Could the number of warnings be printed at the end (together with the number of failed tests, errored tests, and total number of tests) ?

c-cube:
There you go.

Issue #61, opened 2019-04-10
Add Rely backend


bandersongit:
I think making a [Rely](https://github.com/facebookexperimental/reason-native/tree/master/src/rely) backend for qcheck would be a useful. I've seen a few people express interest in creating one, so I thought I would reach out here.

Based on looking at the alcotest and oUnit implementations I think it should be fairly straightforward and I am happy to help with the translation.

c-cube:
I would merge a PR that adds it :slightly_smiling_face: especially as a `qcheck-rely` package.

c-cube:
Possible issue is that esy doesn't seem to be on opam *at all*, which makes a `qcheck-rely` package more difficult, I suppose.

bandersongit:
> Possible issue is that esy doesn't seem to be on opam _at all_, which makes a `qcheck-rely` package more difficult, I suppose.

Definitely. I could try setting up an esy/npm workflow in a subdirectory in order to get the dependencies we need (namely rely), but it would certainly add some maintenance overhead particularly when it comes to releases

c-cube:
if you want `qcheck-rely` (or `rely-qcheck`?) to live in this repo, I'm not opposed to having a package.json or something like that in there for this sub-library. I assume esy is able to work with a repo that mixes opam and esy files. Let me know what you think :slightly_smiling_face:

bandersongit:
> if you want `qcheck-rely` (or `rely-qcheck`?) to live in this repo, I'm not opposed to having a package.json or something like that in there for this sub-library. I assume esy is able to work with a repo that mixes opam and esy files. Let me know what you think 🙂

I'll play with this today and see what I can come up with, thanks for the flexibility!

bandersongit:
Released [@reason-native/qcheck-rely](https://github.com/facebookexperimental/reason-native/tree/master/src/qcheck-rely) to NPM

PR #62, opened 2019-04-15
[WIP] qcheck-rely

bandersongit:
This PR sets up esy as discussed in https://github.com/c-cube/qcheck/issues/61 and provides a potential implementation for qcheck-rely. This implementation was the simplest and I think is a good springboard for discussion.

The advantages of this approach are that it is dead simple to use (see qcheck_rely_test.re), and I think meets the minimum usability bar.

An alternative approach would be to write a [custom matcher](https://reason-native.com/docs/rely/api#using-custom-matchers). This would let us have much more control over the formatting of the output (with the current approach, the output is just the contents of the exception raised by QCheck.Test.check_cell_exn and formatted as a normal exception by Rely) which would let us do things like color the text red and add the random seed in an intuitive place in the output.

The downside of this approach is that I think the experience of actually writing tests becomes worse as rather than
```reason
open TestFramework;
open QCheck_rely;

let myQCheckTest = ...;

describe("my test suite", ({test}) => {
  to_rely(test,   myQCheckTest);
  ...
});
```
usage would become
```reason
open TestFramework;
open QCheck_rely;

/* This syntax will become mandatory in Rely 2.0.0 which will release today */
let { describe } = extendDescribe(QCheck_rely.matchers);

let myQCheckTest = ...;

describe("my test suite", ({test}) => {
  test("my test", ({expect}) => {
    /*.toPass could be anything we want/we could have .toFail etc. as well*/
    expect.ext.qCheck(myQCheckTest).toPass();
  });
  ...
});
```

I also imagine there is a middle ground where we catch the exception within QCheck_rely.to_rely and rethrow with the random seed and whatever non-text coloring adjustments we want (such as removing the redundant stack trace in the exception case) to make from the exception output.

It is also possible to expose both to_rely and custom matchers if we want to support both and let users choose which they'd rather use.

Current output (pending release of https://github.com/facebookexperimental/reason-native/pull/144, which should be soon as I plan on cutting a version of Rely today):
image


bandersongit:
Also, for using esy (assuming you have esy installed, if not go to https://esy.sh/
install dependencies
```
esy install
```
to run the qcheck-rely tests
```
esy example-rely
```

We can definitely release to npm with this configuration, I've not actually published anything to opam so I don't know if further changes would be needed for that. Regardless I imagine the docs will need to be updated as well

src/rely/QCheck_rely.re
```diff
@@ -0,0 +1,42 @@
...
+
+let seed_ =
+  lazy {
+    let s =
+      try (int_of_string @@ Sys.getenv("QCHECK_SEED")) {
+      | _ =>
+        Random.self_init();
+        Random.int(1000000000);
+      };
+    /* Printf.printf("qcheck random seed: %d\n%!", s); */
```
bandersongit: Printing directly messed with the output from Rely so I commented it out. I imagine we will want to have the random seed somewhere and left some thoughts in the description of the PR.

c-cube:
I'm going to go over the PR, thank you very much.
Would it be possible to not commit the huge esy lock directory?

example/rely/TestFramework.re
```diff
@@ -0,0 +1,19 @@
+let projectDir =
```
c-cube: would it be simpler to run tests with an environment variable pointing to the project root?

package.json
```diff
@@ -0,0 +1,20 @@
...
+      "install": [
+        "esy-installer qcheck-rely.install"
+      ]
+    },
+    "scripts": {
+        "example-rely": "dune exec example/rely/QCheck_rely_test.exe"
+    },
+    "dependencies": {
+      "@opam/dune": "*",
+      "@opam/qcheck": "*",
```
c-cube: should the version be `^0.9` there?

src/rely/QCheck_rely.re
c-cube: that makes sense, maybe a local file with the seed inside could be produced, or maybe some form of message (if rely provides such a mechanism)?

bandersongit:
> Would it be possible to not commit the huge esy lock directory?

The esy.lock directory exists to ensure reproducible builds, it is recommended to source control it. I put it in a separate commit to (hopefully) make this easier to review

package.json
bandersongit: ```^0.9``` isn't working because it's getting resolved to ```@opam/qcheck@>=opam:0.9.0 <opam:0.10.``` and 0.9 < 0.9.0. I'm going with ```>= 0.9 < 0.10``` instead. I think this should be good?

c-cube:
I'd still rather not have a lock directory at all (afaict reproducible builds are the problem of binaries, not libraries) because it's *huge*.

bandersongit:
> I'd still rather not have a lock directory at all (afaict reproducible builds are the problem of binaries, not libraries) because it's _huge_.

If you do some googling there's some nice discussion of this topic, but the consensus at least among the makers of package managers is to source control lock files even for libraries, [yarn has a reasonable post summarizing some of the arguments and counter arguments](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/) . The primary arguments for this that I see are
a) Better contributor experience
b) Consistent builds in CI
c) We don't/shouldn't distribute the lockfile with the package itself so the size impact shouldn't be felt by users

That said, it's certainly your call

c-cube:
I'll commit the lock when it's a single file, not a directory (it's on their todolist apparently, I asked on discord). In the meantime I think it's survivable without (perhaps the lockfile can go in a special branch if you really insist) :)

src/rely/QCheck_rely.re
```diff
@@ -0,0 +1,42 @@
...
+  lazy (
+    switch (Sys.getenv("QCHECK_LONG")) {
+    | "1"
+    | "true" => true
+    | _ => false
+    | exception Not_found => false
+    }
+  );
+
+let to_rely =
```
kyldvs: Is there a reason we are not using camelCase for these functions? I feel like we should be keeping Rely conventions for naming since that is the intended audience.

src/rely/QCheck_rely.rei
```diff
@@ -0,0 +1,24 @@
+/**
+  {1 Rely backend for QCheck}
+
+    We can also use environment variables for controlling QCheck here
+
+    - [QCHECK_SEED] if an integer, will fix the seed
+    - [QCHECK_LONG] is present, will trigger long tests
+
+    @since 0.9
```
c-cube: btw this should be `@since NEXT_RELEASE` (or 0.10, worst case)

bandersongit: This brings up the question of what we want to do concerning release? Since Rely is not on opam, I imagine this isn't releasable there without needing to inject some module that would live in Rely. 

I could see releasing this to npm as qcheck-rely, or if you'd rather have it live elsewhere, we could move this to reason-native and release it as @reason-native/qcheck-rely

c-cube: I don't know how reason-native works, is it a kind of giant monorepo? Both options are ok for me, it's your code :slightly_smiling_face: 
If it's simpler to move it to reason-native it's totally fine by me.

src/rely/QCheck_rely.re
bandersongit: I think writing to a temp file raises the same issue and I don't love just writing to the current directory (even only on failure), but could maybe be convinced.

What does a typical workflow look like with qcheck? If there is a failure is running repeatedly with a random seed still likely to produce the failure or is it best practice to hardcode the seed? If you hardcode a random seed when debugging are you reasonably likely to reproduce the failure (if so, how long would that take)?

c-cube: typically the runner displays the seed and provides a flag to set the seed (to one known to trigger some failure). In this case it should always reproduce the same failure if you don't use other sources of randomness. I don't like hardcoding the seed because using a fresh one every time covers more search space over time!

src/rely/QCheck_rely.rei
bandersongit: It's an npm namespace that we are using for a lot of the native reason development tools we are building at Facebook. It also happens to be a monorepo.

I'll think some more about where it makes sense to live.

bandersongit:
I think it will be easiest to release this under reason-native, so I plan on moving this there, but will continue to iterate here for continuity's sake. I camelcased the names to match the conventions we are using in that project.

I spent some time implementing this as custom matchers which leads to better output, but different syntax. Output for this approach is below. I'd appreciate your feedback on the tradeoffs between syntax and quality of output.

image

c-cube:
This looks really good! I think it's the nicest output for qcheck so far (even though our custom runners ain't too bad). Moving to reason-native is probably the simplest, but I'll make a minor remark: you should only have to depend on `qcheck-core`, not `qcheck` (it'd pull ounit which I assume you don't need). godspeed!

bandersongit:
I just released [@reason-native/qcheck-rely](https://github.com/facebookexperimental/reason-native/tree/master/src/qcheck-rely), thanks for building qcheck and for all the help @c-cube! Can we put integration instructions with Rely in the README? (I'm happy to write them)

c-cube:
Congrats! Yes, I'd accept a new section of the README if that helps :)

PR #63, opened 2019-04-17
speed-up list shrinker

jmid:
An improvement to the built-in list shrinker.

### DIFF
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -395,8 +395,45 @@ module Shrink = struct
         done
 
   let list ?shrink l yield =
-    array ?shrink (Array.of_list l)
-      (fun a -> yield (Array.to_list a))
+    let n = List.length l in
+    let chunk_size = ref (n/2) in
+
+    let rec fill_queue n l q = match n,l with
+      | 0,_ -> l
+      | _,x::xs -> (Queue.push x q; fill_queue (n-1) xs q)
+      | _,_ -> failwith "size mismatch in fill_queue"
+    in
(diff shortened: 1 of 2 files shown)


src/core/QCheck.ml
```diff
@@ -1506,10 +1543,11 @@ module Test = struct
     let hist_size, bucket_size =
       let sample_width = Int64.(sub (of_int max_idx) (of_int min_idx)) in
       if sample_width > Int64.of_int stat_max_lines
-      then 1+stat_max_lines,
```
c-cube: does this belong in the other PR?

src/core/QCheck.ml
```diff
@@ -395,8 +395,45 @@ module Shrink = struct
         done
 
   let list ?shrink l yield =
```
c-cube: looks good. I'll refactor the style a bit but the idea is good in any case (similar to integer shrinking I think).

src/core/QCheck.ml
jmid: Hm. That's weird. Indeed it has nothing to do with list shrinking (and was committed a day earlier). I *really* should start using branches... :-)

src/core/QCheck.ml
jmid: Great. Besides the repetitive testing for empty list `[]` the code should mirror the array-based strategy from earlier, as confirmed by matching the exact number of successful shrink steps between the two.

c-cube:
Thank you! :+1:

PR #67, opened 2019-06-21
Gen.fix has a more general and useful type

gasche:


c-cube:
good point, thanks @gasche :)

PR #68, opened 2019-07-12
sequence_list : a t list -> a list t

gasche:
I need to recursively shrink a list of values (I want to enumerate
lists of all shrunk values) and I don't see how to do it without this
function.

Haskell typically exposes both (List specializing the traversable
datatype here)

    sequence : Applicative f => f (List a) -> f (List a)
    traverse : Applicative f => (a -> f b) -> List a -> f (List b)

For now I propose only `sequence` to keep the API simpler, and also
because it is a natural (homogeneous) extension of pair, triple, quad.

### DIFF
--- a/src/core/QCheck.mli
+++ b/src/core/QCheck.mli
@@ -412,11 +412,13 @@ module Iter : sig
 
   val of_list : 'a list -> 'a t
   val of_array : 'a array -> 'a t
+
   val pair : 'a t -> 'b t -> ('a * 'b) t
   val triple : 'a t -> 'b t -> 'c t -> ('a * 'b * 'c) t
   val quad : 'a t -> 'b t -> 'c t -> 'd t -> ('a * 'b * 'c * 'd) t
-  val find : ('a -> bool) -> 'a t -> 'a option
+  val sequence_list : 'a t list -> 'a list t
 
+  val find : ('a -> bool) -> 'a t -> 'a option
   val filter : ('a -> bool) -> 'a t -> 'a t
 
   val append_l : 'a t list -> 'a t
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -318,6 +318,13 @@ module Iter = struct
   let quad a b c d yield =
     a (fun x -> b (fun y -> c (fun z -> d (fun w -> yield (x,y,z,w)))))
 
+  let sequence_list (type a) (gs  : a t list) : a list t = fun yield ->
+    List.fold_left (fun acc g yield ->
+        acc @@ fun rev_xs ->
+        g @@ fun x ->
+        yield (x :: rev_xs)
+    ) (return []) gs @@ fun rev_xs -> yield (List.rev rev_xs)
+
   exception IterExit
   let find_map p iter =
     let r = ref None in


src/core/QCheck.mli
```diff
@@ -412,11 +412,13 @@ module Iter : sig
 
   val of_list : 'a list -> 'a t
   val of_array : 'a array -> 'a t
+
   val pair : 'a t -> 'b t -> ('a * 'b) t
   val triple : 'a t -> 'b t -> 'c t -> ('a * 'b * 'c) t
   val quad : 'a t -> 'b t -> 'c t -> 'd t -> ('a * 'b * 'c * 'd) t
-  val find : ('a -> bool) -> 'a t -> 'a option
+  val sequence_list : 'a t list -> 'a list t
```
c-cube: I think you need to pick the same name :wink:

c-cube:
Seems very useful, but there's a tiny API issue ;-)

gasche:
(I'll fix the naming error.)

Actually @jmid had a different suggestion for an implementation that does not try to shrink all the list elements at once, but only one element of the list each time. His code below:

```ocaml
(* a size-preserving list-shrinker *)
let rec shrink_list_elements ~shrink xs = match xs with
  | [] -> Iter.empty
  | x::xs ->
    Iter.(
      map (fun x' -> x'::xs) (shrink x)
      <+>
      map (fun xs -> x::xs) (shrink_list_elements ~shrink xs))
```

I wonder which approach is the most efficient for shrinking in practice.

With my implementation, the first returned list contain the first shrinking result for each element in the input list. If we assume that the shrinker converges very quickly (after one shrinking we get a correct but unshrinkable element), this is much better than @jmid's implementation which needs as many iterations as the number of elements in the list. On the contrary, if one element cannot be shrunk at all, then merely using my combinator will fail to find any valid result (it will try all choices of shrinking each element and then fail).

Another issue is the empty list: for my need (shrinking a list of shrinkable elements), I would like the empty list to fail (`Iter.empty`) rather than shrinking into the empty list. @jmid's implementation has the right behavior here, and mine does not.

I'm not sure what is best, but my conclusion for now is that a generic `Iter` constructor (at this level I think my implementation is the right/canonical one) has a behavior that is not optimal for the use-case (shrinking a list of values), so this PR is maybe not the best idea. We should rather provide a function in the `Shrink` module, that does something better than the current `Shrink.list`.

Ideally we would have `Shrink.list_spine : 'a list t` and `Shrink.list_elems : 'a t -> 'a list t`, the first one only drops elements of the input and the second only shrinks elements of the input, and then `Shrink.list` is just `compose Shrink.list_spine (Shrink.list_elems ~shrink)`.

c-cube:
There's a similar thing in the `int` shrinker, which tries to both be reasonably "complete" and to try and get to small values if they are valid (by dichotomy). I think a "smart" shrinker that tries a bit of both (try to cut corners first, fallback on the general slower solution) is doable, albeit more complicated.

At a higher level, maybe that's where `qcheck` shows its limitations, and something like Hypothesis, with more advanced combinations of generation and shrinking, start making more sense?

gasche:
What do you think of `Shrink.list_spine` and `Shrink.list_elems`? It's a simple API and it decomposes your (somewhat ugly...) `Shrink.list` combinator.

c-cube:
If we still keep `Shrink.list` as a general purpose shrinker in addition to these new combinators, I see no objection to that.

Issue #64, opened 2019-04-17
List shrinker performance


jmid:
Following up to the revision of the integer shrinker, I somehow
ended up looking into the list shrinker.

First, consider these attempts of the built-in list shrinker:

```
# QCheck_runner.set_seed 158975878;;
# let print_list xs = print_endline (Print.list Print.int xs) in
  Test.check_exn (Test.make (list small_int) (fun xs -> print_list xs; xs = []));;
[9; 73; 794; 4603]
[]
[794; 4603]
[]
[4603]
[]
[2301]
[]
[1150]
[]
[575]
[]
[287]
[]
[143]
[]
[71]
[]
[35]
[]
[17]
[]
[8]
[]
[4]
[]
[2]
[]
[1]
[]
[0]
[]
Exception:
QCheck.Test.Test_fail ("anon_test_13", ["[0] (after 15 shrink steps)"]).
```

For every succesful shrink+restart we repeatedly test whether [] is a
counterexample. Without some form of memorization this is hard to
avoid, but one could argue whether these redundant attempts are
needed. For any false property that happens to hold for the empty
list, half of QCheck's shrinking attempts are in vain.

Secondly, the code for the built-in list shrinker falls back on array
shrinking: 
```
let array ?shrink a yield =
  let n = Array.length a in
  let chunk_size = ref (n/2) in
  while !chunk_size > 0 do
    for i=0 to n - !chunk_size do
      (* remove elements in [i .. i+!chunk_size] *)
      let a' = Array.init (n - !chunk_size)
          (fun j -> if j < i then a.(j) else a.(j + !chunk_size))
      in
      yield a'
    done;
    chunk_size := !chunk_size / 2;
  done;
  match shrink with
  | None -> ()
  | Some f ->
    (* try to shrink each element of the array *)
    for i = 0 to Array.length a - 1 do
      f a.(i) (fun x ->
          let b = Array.copy a in
          b.(i) <- x;
          yield b
        )
    done
    
let list ?shrink l yield =
  array_shk ?shrink (Array.of_list l)
    (fun a -> yield (Array.to_list a))
```

This stresses the garbage collector needlessly by the converting back
and forth to arrays. It could be beneficial to share some of the list
tails instead of allocating a new array repeatedly only to convert it
back to a freshly allocated list afterwards (no sharing guaranteed).

As far as I can tell, this has O(n^2 log n) time complexity: the
repeatedly-halving while-loop runs at most O(log n) times, the
[...]

PR #69, opened 2019-07-12
Shrink: decompose Shrink.list into Shrink.list_spine and Shrink.list_elems

gasche:
See the discussion in #68 -- I think that this PR should supersed #68.

> I'm not sure what is best, but my conclusion for now is that a generic `Iter` combinator (at this level I think my implementation is the right/canonical one) has a behavior that is not optimal for the use-case (shrinking a list of values), so this PR is maybe not the best idea. We should rather provide a function in the `Shrink` module, that does something better than the current `Shrink.list`.
>
> Ideally we would have `Shrink.list_spine : 'a list t` and `Shrink.list_elems : 'a t -> 'a list t`, the first one only drops elements of the input and the second only shrinks elements of the input, and then `Shrink.list` is just `compose Shrink.list_spine (Shrink.list_elems ~shrink)`.

I was a bit surprised to find out that `Shrink.list` is not dropping elements then shrinking the remaining elements recursively, but it is returning `Shrink.list_spine l <+> Shrink.list_elems shrink l`, concatenating the transformations instead of composing them. I did not change the implementation.

### DIFF
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -333,6 +333,7 @@ module Shrink = struct
   type 'a t = 'a -> 'a Iter.t
 
   let nil _ = Iter.empty
+
   let unit = nil
 
   (* balanced shrinker for integers (non-exhaustive) *)
@@ -394,7 +395,7 @@ module Shrink = struct
           )
         done
 
-  let list ?shrink l yield =
+  let list_spine l yield =
     let n = List.length l in
     let chunk_size = ref (n/2) in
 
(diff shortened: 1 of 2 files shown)


c-cube:
That seems nice, I'd just like to have @Gbury 's opinion as well.

jmid:
> I was a bit surprised to find out that `Shrink.list` is not dropping elements then shrinking the remaining elements recursively, but it is returning `Shrink.list_spine l <+> Shrink.list_elems shrink l`, concatenating the transformations instead of composing them. I did not change the implementation.

There's a general setup around shrinking that tries to reduce in small steps after which the shrinker is restarted from the new (smaller) counterexample. With this setup a list shrinker can then first, e.g., cut a list in half (+restart after the success), then continue by, e.g., removing single elements (+restart after the success), and finally get to reducing the remaining individual elements.

I am in favor of extending the API this way, and it would indeed be elegant to express the existing list shrinker as a concatenation of two available API operations.

Please time the resulting shrinker. The list shrinker is also used, e.g., by the function generators. I have spent some time in the past improving its performance - see https://github.com/c-cube/qcheck/issues/64

gasche:
The patch is just a refactoring of the existing code, so there should be absolutely no change in behavior and in performance, just exposing a more finer-grained API. Thanks for the reference to #64 though: I have no idea how to test the performance of shrinkers, and that PR will teach me how to.

Gbury:
The split seems useful indeed, and the code is correct.

While we're at it, I was wondering if we could make the shrinker for lists more complete, by shrinking on multiple elements of the list simultaneously. For instance, the shrinker for the list `[4; 4]` will never yield the list `[2; 2]` (because it shrinks elements individually and separately), which is potentially incomplete[1]. Of coure, given the exponential number of lists to enumerate, this would need to be done after the current shrinking strategy.


[1]: Imagine the following example (maybe a little contrived, but not that much when considering the complexity of some bugs), where the function begin tested fails only when the lists contains consecutive even integers, or even consecutive equal integers (above a certain threshold because of some overzealous optimization/unfolding).

c-cube:
Thanks! I think this is a good first step, more elaborate behaviors (trying to shrink all subsets of the list?) should probably be opt-in and are better discussed in another PR.

Issue #58, opened 2019-01-26
Understanding the implementation of Gen.float


cemerick:
I needed a `float_range` (perhaps a reasonable PR?), and came up with this:

```ocaml
  let float_range a b =
    if b < a then invalid_arg "Gen.float_range";
    let range = b -. a in
    fun st -> a +. RS.float st range
  let (--.) = float_range
```

Seems fine, and appears to work well in my use, but along the way I happened to look at the implementation of `Gen.float`:

https://github.com/c-cube/qcheck/blob/0dab657c3bf30b27592d01248f52149aaa538b2d/src/core/QCheck.ml#L117-L121

Some questions came to mind:

1. Why is this implemented in terms of `exp`, rather than just calling through to `RS.float`?
2. Why not use `RS.bool` to determine whether to negate, rather than generating floats and doing a comparison?
3. Why not use `~-.` instead of performing a multiplication?

c-cube:
This comes from [qtest](https://github.com/vincent-hugot/qtest), it seems. It's old code, I never looked too much at float generators. I suppose it's not that easy to know what range to use, or maybe it's to have a certain shape in the distribution.

A PR for `float_range` would make sense! :+1:

PR #70, opened 2019-07-19
Add float generators with bound and range

ulugbekna:
- add `float_bound_inclusive`, `float_bound_exclusive` generators to generate floats between 0 and given bound (proposed by @gasche)

- add `float_range` to generate floats in the given range (solves #58)

- fix minor indentation inconsistency

Note: QCheck has own float generation algorithm, while I used `Random.State.float` for simplicity.

### DIFF
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -130,6 +130,12 @@ module Gen = struct
     | b_pos when bound > 0. -> RS.float st (b_pos -. epsilon_float)
     | b_neg -> RS.float st (b_neg +. epsilon_float)
 
+  let float_range low high =
+    if high < low || high -. low > max_float then invalid_arg "Gen.float_range";
+    fun st -> low +. (float_bound_inclusive (high -. low) st)
+
+  let (--.) = float_range
+
   let neg_int st = -(nat st)
 
   let opt f st =
@@ -627,6 +633,8 @@ let float_bound_inclusive bound =
 
 let float_bound_exclusive bound =
   make_scalar ~print:string_of_float (Gen.float_bound_exclusive bound)
+
+let float_range low high = make_scalar ~print:string_of_float (Gen.float_range low high)
 
 let int = make_int Gen.int
 let int_bound n = make_int (Gen.int_bound n)
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -122,6 +122,14 @@ module Gen = struct
   let pfloat st = abs_float (float st)
   let nfloat st = -.(pfloat st)
 
+  let float_bound_inclusive bound st = RS.float st bound
+
+  let float_bound_exclusive bound st =
+    match bound with
+    | 0. -> raise (Invalid_argument "Gen.float_bound_exclusive")
+    | b_pos when bound > 0. -> RS.float st (b_pos -. epsilon_float)
+    | b_neg -> RS.float st (b_neg +. epsilon_float)
+
   let neg_int st = -(nat st)
 
   let opt f st =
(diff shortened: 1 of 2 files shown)


src/core/QCheck.mli
```diff
@@ -190,16 +190,33 @@ module Gen : sig
...
+  val bool : bool t (** The boolean generator. *)
 
-  val float: float t   (** Generates floating point numbers. *)
+  val float : float t   (** Generates floating point numbers. *)
 
   val pfloat : float t (** Generates positive floating point numbers (0. included). *)
 
   val nfloat : float t (** Generates negative floating point numbers. (-0. included) *)
 
+  val float_bound_inclusive : float -> float t
```
c-cube: need some `@since NEXT_RELEASE` annotations.

src/core/QCheck.ml
```diff
@@ -122,6 +122,20 @@ module Gen = struct
...
+    match bound with
+    | 0. -> raise (Invalid_argument "Gen.float_bound_exclusive")
+    | b_pos when bound > 0. -> RS.float st (b_pos -. epsilon_float)
+    | b_neg -> RS.float st (b_neg +. epsilon_float)
+
+  let float_range low high =
+    if high < low || high -. low > max_float then invalid_arg "Gen.float_range";
+    fun st -> low +. (float_bound_inclusive (high -. low) st)
+
+  let (--.) = float_range
```
c-cube: I like that! :grin:

src/core/QCheck.mli
```diff
@@ -892,6 +909,19 @@ val pos_float : float arbitrary
 val neg_float : float arbitrary
 (** Negative float generator (no nan and no infinities). *)
 
+val float_bound_inclusive : float -> float arbitrary
```
c-cube: same for `@since NEXT_RELEASE`

src/core/QCheck.mli
```diff
@@ -190,16 +190,33 @@ module Gen : sig
...
+  [bound] is 0, the result is 0. *)
+
+  val float_bound_exclusive : float -> float t
+  (** [float_bound_exclusive bound] returns a random floating-point number between 0 and
+  [bound] (exclusive).  If [bound] is negative, the result is negative or zero.
+  @raise Invalid_argument if [bound] is zero. *)
+
+  val float_range : float -> float -> float t
+  (** [float_range low high] generates floating-point numbers within [low] and [high] (inclusive)
+      @raise Invalid_argument if [high < low] or if the range is larger than [max_float]. *)
```
c-cube: nice doc.

src/core/QCheck.mli
ulugbekna: Should I put `@since 0.10` or literally `@since NEXT_RELEASE`?

c-cube: literally `@since NEXT_RELEASE`, there is a makefile target to update that if you're curious :)

c-cube:
Thanks again! :sparkles:

gasche:
I think it would be interesting to also have `int_bound_{inclusive,exclusive}`. @ulugbekna, is this something that you are also planning to work on?

ulugbekna:
I would be glad to work on `int_bound_{inclusive,exclusive}`. 

I have a couple of questions: 

1) `float_bound_{inclusive, exclusive}` allow to have negative bound; should we consider adding that support to `int_bound_...`? 
Generated ints would be between 0 and `[bound]`, whether bound is positive or negative.

2) if new `int_bound_...` functions are in, should we deprecate `int_bound`? (by putting an indication?)

gasche:
If `int_bound` already supports negative bounds, then I think they should be supported as well. Otherwise I think that supporting negative bound is nice-to-have but not strictly necessary; you should do as you think is best.

I'm not in favor of deprecation generally, it causes more trouble than it's worth. The documentation of `int_bound` could just say "alias of `int_bound_inclusive`" (if that's really the right one), and nobody is hurt by having two functions.

PR #72, opened 2019-07-20
Change implementation of 'int_bound'

ulugbekna:
Together with @gasche, we were implementing our own `Gen.frequency` looking at the implementation by QCheck. 

We saw that to generate a bounded int, it uses `Random.State.int` instead of `int_bound`. As we wanted to have the same values generated by both `Gen.frequency` and our implementation, we also opted for using `Random.State.int`; however, it would be better if we used the API provided by QCheck instead of a lower-level library fn. 

After this change, `int_bound` and `Random.State.int` will generate same values for `bound < 2^{30}` with own implementation for larger values, which would satisfy most of the users, assumably. 

(@gasche can probably phrase it more convincingly if necessary)

### DIFF
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -156,9 +156,9 @@ module Gen = struct
   let int st = if RS.bool st then - (pint st) - 1 else pint st
   let int_bound n =
     if n < 0 then invalid_arg "Gen.int_bound";
-    fun st ->
-      let r = pint st in
-      r mod (n+1)
+    if n <= (1 lsl 30) - 1
+    then fun st -> Random.State.int st n
+    else fun st -> let r = pint st in r mod (n + 1)
   let int_range a b =
     if b < a then invalid_arg "Gen.int_range";
     fun st -> a + (int_bound (b-a) st)
--- a/src/core/QCheck.mli
+++ b/src/core/QCheck.mli
@@ -243,6 +243,7 @@ module Gen : sig
 
   val int_bound : int -> int t
   (** Uniform integer generator producing integers within [0... bound].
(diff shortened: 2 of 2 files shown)


gasche:
The general context we were in is that we were trying to improve the efficiency of one of our generators; instead of a relatively naive implementation built on top of the existing QCheck primitives, we wanted to use a specialized implementation. The two implementations use the same amount of entropy (it's about picking a value in the same value space, just with less computations).

It's very useful for testing if we can get the two implementations to consume the random-generator state in the exact same way; it means that the generation behavior of the two versions will be identical, enabling meaningful performance comparisons. This property depends on how QCheck primitives consume the state internally. For most of them there is an "obvious" way to do it (picking among a list probably does an `int_bound (list_size li - 1)`, picking among a weighted list picks among the sums of weights, etc.); it would be good if the implementation actually behaved in this "obvious" way and not in a different way.

For int-consuming combinators (most of them), this principle is currently broken because `int_bound` and `Random.State.int` consume the random-generator state differently. This PR by @ulugbekna fixes the discrepancy in essentially all cases in prcatice.

src/core/QCheck.ml
```diff
@@ -154,11 +154,11 @@ module Gen = struct
         lor ((RS.bits st land 3) lsl 60)  (* Top 2 bits *)  (* top bit = 0 *)
 
   let int st = if RS.bool st then - (pint st) - 1 else pint st
-  let int_bound n =
+  let int_bound n st =
     if n < 0 then invalid_arg "Gen.int_bound";
```
c-cube: is there a reason to not check that anymore before `st` is passed?

ulugbekna: Nope, unintended omission. Will fix and force push.

PR #73, opened 2019-07-20
add weighted shuffled lists generator

ulugbekna:
Adds a function `shuffle_w_l`, which creates a generator that can shuffle the given list according to the elements' weights.

### DIFF
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -209,6 +209,14 @@ module Gen = struct
     shuffle_a a st;
     Array.to_list a
 
+  let shuffle_w_l l st =
+    let sample (w, v) =
+      let fl_w = float_of_int w in
+      (float_bound_inclusive 1. st ** (1. /. fl_w), v)
+    in
+    let samples = List.map sample l in
+    List.sort (fun (w1, _) (w2, _) -> compare w2 w1) samples |> List.map snd
+
   let pair g1 g2 st = (g1 st, g2 st)
 
   let triple g1 g2 g3 st = (g1 st, g2 st, g3 st)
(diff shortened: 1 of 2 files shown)


src/core/QCheck.mli
```diff
@@ -190,6 +190,9 @@ module Gen : sig
   val shuffle_l : 'a list -> 'a list t
   (** Creates a generator of shuffled lists. *)
 
+  val shuffle_w_l : (int * 'a) list -> 'a list t
+  (** Creates a generator of weighted shuffled lists. *)
```
c-cube: this needs a `@since` annotation and, I'm afraid, a lot more explanations :)
An example or two would be very useful (possibly also the intended use case).

src/core/QCheck.ml
```diff
@@ -209,6 +209,14 @@ module Gen = struct
     shuffle_a a st;
     Array.to_list a
 
+  let shuffle_w_l l st =
+    let sample (w, v) =
+      let fl_w = float_of_int w in
+      (float_bound_inclusive 1. st ** (1. /. fl_w), v)
+    in
+    let samples = List.map sample l in
```
c-cube: you could use `List.rev_map` since it's going to be sorted afterwards (and possibly sort it backwards to use `List.rev_map snd`).

c-cube:
This seems useful, thank you. Just a few minor nitpicks.
Btw would you and @gasche like to get commit access?

src/core/QCheck.mli
ulugbekna: Sure, that's a great idea! 
Had the same feeling.

src/core/QCheck.ml
ulugbekna: That's a good idea.

ulugbekna:
Re access: 
I would personally prefer my work to be reviewed before it's merged, if I understood you correctly.

PR #74, opened 2019-07-22
fix 'int_bound' bound inclusiveness problem

ulugbekna:
When I changed `int_bound` to use `Random.State.int` (`RS.int`), I did not take into consideration that `int_bound` generates values including `bound`, while `RS.int` doesn't. 

This PR fixes that omission. (It also changes the max bound for using `RS.int` from `2^{30}-1` to `2^{30}-2`, so that we can pass `bound+1` to `RS.int` without int representation problems on 32-bit machines.


c-cube:
Good point, thanks!

PR #77, opened 2019-09-28
Add QCheck.Gen.{string_of,string_readable}

### LINKED ISSUE #75: Optional labelled parameter in functions
I've come across a couple of functions in [`QCheck.Gen`](https://c-cube.github.io/qcheck/0.9/qcheck-core/QCheck/Gen/index.html) which have only optional parameters:

```ocaml
val string : ?⁠gen:char t ‑> string t
val small_string : ?⁠gen:char t ‑> string t
```

Not sure if there are more throughout the library. Functions with only optional labelled parameters are tricky to use :-) would there be some receptiveness to adding a 'bookend' `()` parameter to them for 1.0? If we're going by SemVer, backward-incompatible changes are allowed in a major version...

yawaramin:
These are more explicit versions of `QCheck.Gen.string` without optional
parameters.

Fixes #75


yawaramin:
@c-cube I added a deprecation attribute to `QCheck.Gen.string` although I see you don't use those in the codebase, I thought it would be a good move to use the 'official' way but can removed if not desired. Also I can update the new doc comments with a `@since` marker but I'm not sure what to use (`NEXT_RELEASE` or `0.11`)

yawaramin:
No problem, I'll make the changes.

c-cube:
Perfect, thank you! :grin:

PR #88, opened 2019-11-23
Add debug shrinking log

### LINKED ISSUE #71: What's the workflow to observe/debug shrinking choices? (with proposals inside)
Here is my current scenario.

I am using a known-bug case to improve my shrinkers. I run the QCheck tests from the command-line, at the fixed seed that is known to run into the bug, and I look at the shrunk counter-example that is displayed. It doesn't look very minimal to me, so I tweak the shrinkers to get a better counter-example. When that works, I get a better counterexample and I iterate. When that does not work, I would like to understand why, in particular to understand the shrinking choices that were made.

So: I would like to use the QCheck API to debug shrinking. In particular, I would like to print (to stderr or a log file) every successful shrinking step (not all shrinking steps, which I can do by adding a debug print to my shrinker, but is too verbose). How can I do this with QCheck?

It looks like what I want is to pass my own `'a Test.handler` value, which observes all events, and add a debug print on the `Shrunk` event. But how can I pass my own handler using the API? As far as I can tell, there is no easy way to do this. `QCheck.Test.check_cell` accepts a handler as parameter, but there is no call to `check_cell` in my code, I am only using `QCheck_runner.run_tests_main`. This function internally calls `check_cell`, but there is no way for the user to pass a custom `handler` parameter to that call.

I started building a PR to do this, but then I noticed that, at this point in the test pipeline, the type of counterexamples is an existential variable (packed in the `Test` or `TestResult` constructor). Not only would an additional parameter to `run_tests_main` need to provide a polymorphic handler (noise and type complexity incoming), it's simply not the right place to add an event handler -- I want to do this while I still know what the type of my counterexample is.

In my code, the type information is known at the callsite of `QCheck.Test.make`. I could add an extra `?handler` parameter to this function, or I could easily turn my call into a lower-level call to `Test.make_cell` and add an extra `?handler` parameter there.

The problem is that cells do not carry handlers with the current API, cell *checkers* do. (And they, again, in the current pipeline, are called after the type parameter of the cell was forgotten.) So I thought I would add a `handler` field to the cells themselves; at checker-creation time, `check_cell` would compose the existing handlers (if any) with its own handler (if any). (The same reasoning might work for `step` and `callback`, but let's put this aside for now.)

The problem is that the handler type is defined as follows:

```ocaml
type 'a handler = string -> 'a cell -> 'a event -> unit
```

the handler gets passed the cell itself. If the handler needs access to the cell, and we decide that the cell should be able to carry handler, we have a cycle in type dependencies. It can be made to work, but it's ugly/complex, and it suggests that it is not the right design.

Why do handlers, steppers and callbacks have access to the cell itself? I don't see anything of value in the `'a cell` record that I would like to have access to in my handler (except maybe the `law` and `arbitrary` fields). In fact, the handler defined in `QCheck_runner` (which I suspect is the only one most QCheck users can use, due to the difficulty in hooking at the `check_cell` cell) exactly ignores the cell parameter.

Is there a better way to solve my need (debugging the shrinker behaviors on a specific CLI run) that I haven't thought about?

In the short term, I might try to send a PR using a dependency cycle between cells and handlers/steppers/callbacks; it's ugly and I think it's the wrong design, but it's non-invasive and backward-compatible.

In the medium term, could we consider dropping the `'a cell` parameter from handlers? (Is it reasonable for steppers and callbacks to do the same? I don't know what typical use-cases for those are.)? Should we pass nothing to replace it, or pass the law and arbitrary? Is this breakage of API something that QCheck developers are okay with?

In the long term, maybe there is a way to break the default runner, a big monolithic beast, so that this kind of flexibility can be obtained by some other approach than "let's add three more optional parameters to this function"? As a user, here is the experience I would like to have: given a specific CLI invocation of my tester, I would like to replace the call to `run_tests_main` in my program by a simple composition of calls of a lower-level API that (1) initializes various state in the same way, so that the program behavior is reproduced, (2) gives me more flexibility to operate on my tests, cells, checkers etc., and (3) gives me the opportunity to report the behavior in the same way, so that the interface doesn't change too much. In short, I'm suggesting that it would be useful to invert the control of `run_test_main`, so that I can access its easy-initialization and easy-reporting facilities, but be the one driving the logic in the middle.

Gbury:
Should fix  #71 

cc @gasche , sorry for the delay.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ example-ounit-test:
 	@dune exec example/ounit/QCheck_ounit_test.exe
 
 example-runner:
-	@dune exec example/QCheck_runner_test.exe
+	@dune exec example/QCheck_runner_test.exe -- --debug-shrink=log.tmp
 
 example-alcotest:
 	@dune exec example/alcotest/QCheck_alcotest_test.exe
--- a/src/runner/QCheck_base_runner.mli
+++ b/src/runner/QCheck_base_runner.mli
@@ -61,7 +61,7 @@ val set_time_between_msg : float -> unit
 
 (** {2 Event handlers} *)
 
-type counter = {
+type counter = private {
   start : float;
   expected : int;
   mutable gen : int;
--- a/src/runner/QCheck_base_runner.mli
+++ b/src/runner/QCheck_base_runner.mli
@@ -58,10 +58,43 @@ val set_time_between_msg : float -> unit
 (** Set the minimum tiem between messages.
     @since 0.9 *)
 
+
+(** {2 Event handlers} *)
+
+type counter = {
+  start : float;
+  expected : int;
+  mutable gen : int;
+  mutable passed : int;
+  mutable failed : int;
+  mutable errored : int;
+}
+(** The type of counter used to keep tracks of the events received for a given
+    test cell. *)
+
+type handler = {
+  handler : 'a. 'a QCheck.Test.handler;
+}
+(** A type to represent polymorphic-enough handlers for test cells. *)
+
+type handler_gen =
+  colors:bool ->
+  debug_shrink:(out_channel option) ->
+  debug_shrink_list:(string list) ->
+  size:int -> out:out_channel -> verbose:bool -> counter -> handler
+(** An alias type to a generator of handlers for test cells. *)
+
+val default_handler : handler_gen
+(** The default handler used. *)
+
+
 (** {2 Run a Suite of Tests and Get Results} *)
 
 val run_tests :
+  ?handler:handler_gen ->
   ?colors:bool -> ?verbose:bool -> ?long:bool ->
+  ?debug_shrink:(out_channel option) ->
+  ?debug_shrink_list:(string list) ->
   ?out:out_channel -> ?rand:Random.State.t ->
   QCheck.Test.t list -> int
 (** Run a suite of tests, and print its results. This is an heritage from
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -20,7 +20,7 @@ example-ounit-test:
 	@dune exec example/ounit/QCheck_ounit_test.exe
 
 example-runner:
-	@dune exec example/QCheck_runner_test.exe
+	@dune exec example/QCheck_runner_test.exe -- --debug-shrink=log.tmp
```
c-cube: this is more a topic for discussion, not a review, but here `Sys.getenv` might also be quite convenient compared to passing parameters on the CLI (which needs to be negociated with each runner, including alcotest).

src/runner/QCheck_base_runner.mli
```diff
@@ -58,10 +58,43 @@ val set_time_between_msg : float -> unit
 (** Set the minimum tiem between messages.
     @since 0.9 *)
 
+
+(** {2 Event handlers} *)
+
+type counter = {
```
c-cube: could this be a `type counter = private {…` to prevent mutation of the fields?

Gbury: That would be a good idea, yes.

jmid:
> Should fix #71

First of all: Nice work! 👍🏻
I've recently spent some time debugging a non-trivial shrinker. For that I found logging each successful shrink attempt to a separate file (`shrink01`, `shrink02`, ...) and using the 2-column output of `diff -y` worked very well to visualize and understand the gradual shrinking process.

gasche:
I just helped a student work on a shrinker for [Inferno](https://gitlab.inria.fr/fpottier/inferno), and we benefited enormously of the new logging feature. Thanks!

PR #95, opened 2020-06-20
modify int_range to make it accept ranges bigger than max_int.

ghilesZ:
### This PR aims at removing the restriction that makes `int_range` fail on ranges bigger than `max_int`.
The new version first checks if the given intertval has a constant sign, in which case its range is smaller than `max_int`.
Otherwise, it splits it to have two constant intervals and then chooses uniforml (i.e. according to their size ratios computed in float) the one in which he branches to `int_bound`


c-cube:
Brilliant. Thank you.

PR #109, opened 2021-04-12
Feat/integrated shrinking

Sir4ur0n:
Closes https://github.com/c-cube/qcheck/issues/106

This is a WIP MR, many things are still pending, I will try to list them below

- [x] Add list shrinking (including `list_towards` stuff)
- [x] Add function shrinking: tracked in https://github.com/c-cube/qcheck/issues/110
- [x] Provide a mean (variant type?) to use the old or new QCheck. See discussion at https://github.com/c-cube/qcheck/pull/109#discussion_r611807647
- [x] Give it a try and check that `Gen.t` is lazy enough
- [x] Check off-by-1 errors e.g. in `int_bound`
- [x] Change `shuffle_a` to make a copy instead of mutating?
- [x] Should we keep `add_shrink_invariant`? If yes: maybe this should be done in another MR to add cleaner post-filtering
- [x] Should we keep `map_keep_input`? IMHO it becomes useless (commented out)
- [x] Generate `mli`
- [x] Reformat
- [x] Check / improve documentation (migration guide?)
- [x] Remove unused code

PR #114, opened 2021-05-09
`nat_split`, `pos_split`

gasche:
The goal of this PR is to implement a function that takes a natural number `n` and parameter `k`, and uniformly splits `n` into `k` numbers `n1..nk` that sum back to `n` (`n1 + ... +nk = n`). This is useful when splitting "fuel" or "size" in a random generator, among `k` components instead of just two. (cc @olivier-martinot, with whom I am collaborating on a random generator that would benefit.)

Two variants are provided: in `nat_split`, the `n1..nk` are natural numbers (range `[0;n]`), in `pos_split` they are strictly positive (range `[1;n]`, or in fact `[1;n-k]`).

The code for these functions is simple and relies on a generic helper that is also useful. However getting them right is tricky, and random generators are hard to test. I checked the implementation in an experimental version of my [random-generator](https://github.com/gasche/random-generator/) library that computes finite distributions.

```
# open Random_generator.Generator.Make(Random_generator.Prob_monad.Distr);;
# Gen.Nat.split ~size:3 5 |> Gen.run ~cmp:Stdlib.compare;;
- : ((int * int) * int list) list =
[((1, 21), [0; 0; 5]); ((1, 21), [0; 1; 4]); ((1, 21), [0; 2; 3]);
 ((1, 21), [0; 3; 2]); ((1, 21), [0; 4; 1]); ((1, 21), [0; 5; 0]);
 ((1, 21), [1; 0; 4]); ((1, 21), [1; 1; 3]); ((1, 21), [1; 2; 2]);
 ((1, 21), [1; 3; 1]); ((1, 21), [1; 4; 0]); ((1, 21), [2; 0; 3]);
 ((1, 21), [2; 1; 2]); ((1, 21), [2; 2; 1]); ((1, 21), [2; 3; 0]);
 ((1, 21), [3; 0; 2]); ((1, 21), [3; 1; 1]); ((1, 21), [3; 2; 0]);
 ((1, 21), [4; 0; 1]); ((1, 21), [4; 1; 0]); ((1, 21), [5; 0; 0])]

utop # Gen.Nat.pos_split ~size:3 5 |> Gen.run ~cmp:Stdlib.compare;;
- : ((int * int) * int list) list =
[((1, 6), [1; 1; 3]); ((1, 6), [1; 2; 2]); ((1, 6), [1; 3; 1]);
 ((1, 6), [2; 1; 2]); ((1, 6), [2; 2; 1]); ((1, 6), [3; 1; 1])]
```

Apparently there are `21` different ways to split `5` into three natural numbers, and `6` ways to split it into three strictly-positive numbers. (The pair `(a, b)` before an output represents the rational number `a/b`, the probability of observing this output.)

### DIFF
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -328,6 +328,36 @@ module Gen = struct
     let rec f' n st = f f' n st in
     f'
 
+  (* nat splitting *)
+
+  let nat_split2 n st =
+    if (n < 2) then invalid_arg "nat_split2";
+    let n1 = int_range 1 (n - 1) st in
+    (n1, n - n1)
+
+  let pos_split2 n st =
+    let n1 = int_range 0 n st in
+    (n1, n - n1)
+
+  let pos_split ~size:k n st =
+    if (k > n) then invalid_arg "nat_split";
+    (* To split n into n{0}+n{1}+..+n{k-1}, we draw distinct "boundaries"
+       b{-1}..b{k-1}, with b{-1}=0 and b{k-1} = n
+       and the k-1 intermediate boundaries b{0}..b{k-2}
+       chosen randomly distinct in [1;n-1].
+
+       Then each n{i} is defined as b{i}-b{i-1}. *)
+    let b = range_subset ~size:(k-1) 1 (n - 1) st in
+    Array.init k (fun i ->
+      if i = 0 then b.(0)
+      else if i = k-1 then n - b.(i-1)
+      else b.(i) - b.(i-1)
+    )
+
+  let nat_split ~size:k n st =
+    pos_split ~size:k (n+k) st
+    |> Array.map (fun v -> v - 1)
+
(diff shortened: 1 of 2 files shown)


c-cube:
That's a good idea. If I recall correctly you use this in recursive sized generators, to allocate size "tokens" between recursive calls, correct?

I wonder about the interaction with #109 . It seems like this shouldn't "shrink" in any meaningful way (apart from the fact it's probably used with `>>=` in most cases), so minimal interaction?

gasche:
> I wonder about the interaction with #109 . It seems like this shouldn't "shrink" in any meaningful way (apart from the fact it's probably used with `>>=` in most cases), so minimal interaction?

Yes, a common way to use this (but not the only way) would be to first pick a random number `k`, then split the current size/fuel k-ways. The split itself cannot be shrunk, but you can re-split after shrinking `k`.

PR #118, opened 2021-05-26
Option generator with an optional ratio

### LINKED ISSUE #117: Option generator
I'm sorry If that question has already been asked.

Is there a particular reason `Gen.opt` produces 15% of `None` ?

Could we (or I):
* Add comments about that choice
* Handle % of `None` as function optional parameter

vch9:
Closes #117 

This MR contains two commits:
* The optional argument on both `Generator` and `Arbitrary` opt
* Alcotest tests

The testing part is a bit tricky since it's probabilistic, but I didn't manage to make the test failed.
The range of possible value might be enough.


src/core/QCheck.mli
```diff
@@ -334,7 +334,8 @@ module Gen : sig
   val array_repeat : int -> 'a t -> 'a array t
   (** [array_repeat i g] builds an array generator from exactly [i] elements generated by [g]. *)
 
-  val opt : 'a t -> 'a option t (** An option generator. *)
+  val opt : ?ratio:float -> 'a t -> 'a option t
+  (**  An option generator, with optional ratio of Some. *)
```
c-cube, line 338:
```suggestion
  (**  An option generator, with optional ratio.
      @param ratio a float between [0.] and [1.] indicating the probability of a sample to be [Some _]
      rather than [None]. This parameter is @since NEXT_RELEASE .
  *)
```

src/core/QCheck.ml
```diff
@@ -139,9 +139,9 @@ module Gen = struct
 
   let neg_int st = -(nat st)
 
-  let opt f st =
+  let opt ?(ratio = 0.75) f st =
```
c-cube, line 142:
```suggestion
  let opt ?(ratio = 0.85) f st =
```

c-cube:
Thank  you, that seems nice. I put the old value back ;)

vch9:
Yep thank you about the change. I must have been tired when I did 0.75 + 0.15 = 1

Issue #112, opened 2021-04-28
Move to Ocamlformat?


Sir4ur0n:
As formatting is a matter of taste, I am opening this issue because I massively prefer Ocamlformat over ocp-indent (which - I think ? - QCheck code currently uses) :grin: 

What do you think about moving to Ocamlformat? I find the formatting much more consistent and helpful to read code

c-cube:
If we find an ok style, why not :)

vch9:
I strongly agree with @Sir4ur0n. It also makes win a LOT of time, got pretty sad when I disabled it in this project

c-cube:
I'm ok with it if we can find a stable config that does *not* put `;` first in records. Should be done in between PRs though, because it makes merging/rebasing a nightmare.

PR #128, opened 2021-06-11
WIP: Proposal of ocamlformat

vch9:
I took the .ocamlformat used in Tezos. I quiet like it, I applied it
on a file so you can take a look

Related to #112


src/core/QCheck.ml
```diff
@@ -1350,17 +1528,15 @@ let find_example ?(name="<example>") ?count ~f g : _ Gen.t =
       Test.make_cell ~max_fail:1 ?count arb (fun x -> not (f x))
     in
     let res = QCheck2.Test.check_cell ~rand:st cell in
-    begin match QCheck2.TestResult.get_state res with
```
c-cube, line 1353: oh for the love of god, WHY

src/core/QCheck.ml
```diff
@@ -883,317 +1026,322 @@ let map ?rev f a =
...
     let buf = Buffer.create 32 in
-    let rec aux
-      : type f. Buffer.t -> f fun_repr -> unit
-      = fun buf r -> match r with
-        | Fun_map (_, r') -> aux buf r'
-        | Fun_tbl r ->
-          Buffer.add_string buf (Poly_tbl.print r.fun_tbl);
-          Printf.bprintf buf "_ -> %s" (match r.fun_arb.print with
+    let rec aux : type f. Buffer.t -> f fun_repr -> unit =
+     fun buf r ->
```
c-cube, line 1219: one space? weird

c-cube (commented):
It's not bad, but I really wish ocamlformat would stop messing with begin/end and my parenthesis in `if then else`. They're here for a reason :angry:

c-cube:
It seems like there's an option `ext-grouping=preserve` that could fix my beef with () and begin/end?

vch9:
> It seems like there's an option `exp-grouping=preserve` that could fix my beef with () and begin/end?

From the manual point of view yes, but I tried on the branch and it always remove the begin/end with `preserve`

Sir4ur0n:
> They're here for a reason 😠

This is a genuine question: what's the reason? 😅 I don't see much the point of useless parentheses except in very specific cases (mostly, to make explicit AND/OR logic in boolean expressions).

c-cube:
They improve readability and are more robust in the face of change. In particular, () if in and begin/and around match makes it less likely to be bitten by OCaml's syntax flaws (if one nests pattern matches or adds `;` in a if branch or after a match).

src/core/.ocamlformat
```diff
@@ -0,0 +1,10 @@
+version=0.18.0
+wrap-fun-args=false
+let-binding-spacing=compact
+field-space=loose
+sequence-style=separator
+doc-comments=before
+margin=80
+module-item-spacing=sparse
+parens-tuple=always
+parens-tuple-patterns=always
```
Sir4ur0n, line 10: Please configure your editor/Git to always put an empty line at the end of all files 😅 See e.g. https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file

vch9: My editor is supposed to do it, I must have removed it by mistake :/

Sir4ur0n:
If `exp-grouping=preserve` preserves brackets, would you be ok with replacing all `begin end` blocks with brackets, @c-cube ? The lack of formatter drives me a bit crazy 😅 If you're ok with this, I don't mind preparing the MR today

c-cube:
I think it preserves nothing, and that's really annoying, the style is worse for it.

Sir4ur0n:
Apparently it replaces `begin end` with brackets, according to https://github.com/ocaml-ppx/ocamlformat/issues/1684

c-cube:
ugh

c-cube:
ocaml's grammar really wasn't designed for a formatter…

Sir4ur0n:
See https://github.com/c-cube/qcheck/pull/147 the `begin end` block of `make_` is replaced with brackets

vch9:
Otherwise there's `[@@ocamlformat "disable"]`..

vch9:
`make build @fmt` applies ocamlformat everywhere, we could use its status in the CI

example/.ocamlformat
```diff
@@ -0,0 +1,10 @@
+version=0.18.0
+wrap-fun-args=false
+let-binding-spacing=compact
+field-space=loose
+sequence-style=separator
+doc-comments=before
+margin=80
+module-item-spacing=sparse
+parens-tuple=always
+parens-tuple-patterns=always
```
Sir4ur0n, line 10: Isn't `exp-grouping=preserve` missing? (don't forget to reformat from `master` after applying this change)

example/.ocamlformat
```diff
@@ -0,0 +1,10 @@
+version=0.18.0
```
Sir4ur0n, line 1: Can't we put a single `.ocamlformat` at the root of QCheck directory, rather than duplicating this file? According to https://github.com/ocaml-ppx/ocamlformat#options this should be enough, right?

Sir4ur0n (changes requested):

example/.ocamlformat
vch9: Done

example/.ocamlformat
vch9: Done

example/.ocamlformat
Sir4ur0n: I don't have the permission to resolve my own threads, but now it's ok for me :+1:

example/.ocamlformat
Sir4ur0n: I don't have the permission to resolve my own threads, but now it's ok for me :+1:

Sir4ur0n:
I suppose this PR is no longer WIP, right?

@c-cube what do you think about this PR?

c-cube:
well for a start I can't even use it locally because it requires a super recent ocamlformat which requires dune 2.8 :roll_eyes:

Sir4ur0n:
I just want to make one thing clear: I'm not particularly invested in Ocamlformat (though it's the one i use everyday on my other project). **What I do wish** is a formatter, any formatter, so that I no longer need to bother with formatting when hacking QCheck 😅 As far as I'm concerned, Ocp-indent or any other tool is also fine (though Ocamlformat and Ocp-indent both integrate with Emacs so I slightly prefer those 😄 )

c-cube:
you can use ocp-indent, in fact I'm surprised I never added a config file for it. However it is not really a formatter since it only does indentation?

c-cube:
Also of course: I don't know of any other formatters for OCaml, so your statement doesn't change much, @Sir4ur0n :)

vch9:
> edit: I sound grumpy. I think this PR will go through — even though I really dislike how ocamlformat breaks my code — because it helps y'all newer contributors.

This is still your project after all, but I won't lie, ocamlformat does save me a lot of time.

vch9:
Looks like we can save `begin and` keywords: https://github.com/ocaml-ppx/ocamlformat/releases/tag/0.19.0

c-cube:
Awesome. Please update, this is exciting now :)

vch9:
Should `"ocamlformat" {>= "0.19.0}"` be added to every `.opam` files at root/?
I have never used a project with multiple packages, I don't know how that kind of external dependency should be declared.

c-cube:
Absolutely no dependency. The code doesn't need ocamlformat, even developping shouldn't _strictly_ need it.

vch9:
> 
> 
>  even developping shouldn't _strictly_ need it.

Depends on whether we want to add formatting in the CI or not

c-cube:
> Depends on whether we want to add formatting in the CI or not

that's true. That's a good argument for having it be a warning somehow (well CI fails is a warning) but not an obstable to pushing. Formatting shouldn't get in the way of, say, bugfixing (and a new ocamlformat that breaks style shouldn't break CI…)

src/core/QCheck2.ml
```diff
@@ -1397,104 +1543,109 @@ module Test = struct
...
-           iter_find_map
-             (fun x ->
-               (* let Tree.Tree (x, _) = x_tree in *)
-               try
-                 incr count;
-                 st.handler st.test.name st.test (Shrinking (steps, !count, x));
-                 begin match run_law st.test.law x with
-                 | Run_fail m when not is_err -> Some (Tree.pure x, Shrink_fail, m)
-                 | _ -> None
-                 end
```
vch9, line 1442: Looks like not every `begin end` is preserved.

vch9: From what I understand, if a `begin end` statement is already in a group it is not preserved.

c-cube: back to the beginning then… :/

c-cube: I wish ocamlformat had a mode where it would just *not* touch parentheses. Let the programmer decide.

PR #131, opened 2021-06-12
Remove arbitrary from QCheck2

### LINKED ISSUE #119: Remove `arbitrary` and use `Gen.t` directly in tests
Copy-pasting and shaving off [my comment in QCheck2 MR](https://github.com/c-cube/qcheck/pull/109#issuecomment-828611466):

-----------------------------------------------

Does it still make sense to keep `arbitrary` (note: Once QCheck2 is merged)? I feel like we would be just as happy with only `Gen.t` (no matter if we keep it here, or if `arbitrary` becomes `Gen.t`).

On top of `Gen.t`, `arbitrary` has:
* pretty printing
* stats
* collect

As @jmid interestingly [pointed out](https://github.com/c-cube/qcheck/pull/109#pullrequestreview-635419995), both stats and tag collection could be moved out as their usage is a bit clunky. Maybe some kind of wrapper so that most users don't even see them, and those who do only wrap in the few places they are interested in stats.

Which leaves us with the pretty printer.

My gut feeling is: we could put the pretty printer in `Gen.t` (as long as it remains optional) and add a `val with_printer : printer -> 'a Gen.t -> 'a Gen.t` function.

The documented pattern, and probably the most used one, would be `|> with_printer foo_printer`:
```ocaml
let foo_gen : foo Gen.t = 
  a_gen >>= fun a ->
  b_gen >>= fun b ->
  make_foo a b
  |> with_printer Foo.pp
```

Pros:
* users now only manipulate one main QCheck type, which is a big UX and verbosity win
* users usually want to combine generators, and pass the pretty printer at the end (just before passing a generator to a test). Intermediary generators only used for composition don't need a pretty printer
* Take the opportunity of this breaking change to fix some deprecations, errors/mistakes (like `ui32`), make the API consistent

Cons:
* This is undoubtedly a breaking change, and a big one :see_no_evil: 
* Needs significant work

Would that be a QCheck3? What is the versioning strategy if we go the way of a full breaking change?

The more I write QCheck code, the more this juggling between `Gen.t` and `arbitrary` is off-putting (and hard to understand its need as a user, as anyway print/collect/stats are optional)

Note: It may be wise to do https://github.com/c-cube/qcheck/issues/107 together with this issue, as this would be highly conflictual and/or involve a lot of throw-away code to do either one first

vch9:
`'a arbitrary` is totally removed.

There is still TODOs inside the code, some can be fixed either by me but I need to fetch some information or
you can suggest the changes (it's mainly documentation).

print, stats and collect are inside `'a cell`:
```ocaml

  type 'a cell = {
    count : int; (* number of tests to do *)
    long_factor : int; (* multiplicative factor for long test count *)
    max_gen : int; (* max number of instances to generate (>= count) *)
    max_fail : int; (* max number of failures *)
    law : 'a -> bool; (* the law to check *)
    gen : 'a Gen.t; (* how to generate/shrink instances *)
    print : 'a Print.t option; (* how to print values *)
    collect : ('a -> string) option; (* /!\ TODO: collect ? *)
    stats : 'a stat list; (* distribution of values of type 'a *)
    qcheck1_shrink : ('a -> ('a -> unit) -> unit) option; (* QCheck1-backward-compatible shrinking *)
    if_assumptions_fail: [`Fatal | `Warning] * float;
    mutable name : string; (* name of the law *)
  }
```
Retro-compatibility with QCheck1 works thanks to @Sir4ur0n work in QCheck1:
```ocaml
  let make_cell ?if_assumptions_fail
      ?count ?long_factor ?max_gen
  ?max_fail ?small:_removed_in_qcheck_2 ?name arb law
  =
  let {gen; shrink; print; collect; stats; _} = arb in
  QCheck2.Test.make_cell_from_QCheck1 ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?name ~gen ?shrink ?print ?collect ~stats law

  let make ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?name arb law =
    QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?name arb law)
```

The `print` move from `arbitrary` to `cell` has one impact: `Observable`. Currently you can't not provide any printing for
an observable, but I believe it's doable with some workaround the type.

Actions:
- [X] Remove `'a arbitrary` from QCheck2
- [X] Retro-compatibility from QCheck1
- [x] Remove every TODO
- [x] Print observable
- [ ] Tests on retro-compatibility (compare `QCheck2.Test.t` from QCheck1 and QCheck2 

Closes #119

### DIFF
--- a/src/core/QCheck2.ml
+++ b/src/core/QCheck2.ml
@@ -1,6 +1,7 @@
 (*
 QCheck: Random testing for OCaml
-copyright (c) 2013-2017, Guillaume Bury, Simon Cruanes, Vincent Hugot, Jan Midtgaard, Julien Debon
+copyright (c) 2013-2017, Guillaume Bury, Simon Cruanes, Vincent Hugot,
+Jan Midtgaard, Julien Debon, Valentin Chaboche
 all rights reserved.
 *)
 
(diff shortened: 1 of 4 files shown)


src/core/QCheck2.ml
```diff
@@ -1099,13 +934,15 @@ end = struct
           with Not_found ->
             if extend then (
               (* Generate a new value and "record" the binding for potential future display/shrinking *)
-              let value_tree = v_arb.gen st in
+              let value_tree = v_gen st in
               p_tree_bindings_rev := (key, value_tree) :: !p_tree_bindings_rev;
               let v = Tree.root value_tree in
               T.add tbl key v;
               Some v
             ) else None) in
```
c-cube, line 942: maybe the whole `create` could take an optional printer? Or even more modular, the function to print a table could take a printer for the key and value types, as it is done in the stdlib and elsewhere.

vch9: I prefer the optional printer in `create` as it is used in `print_repr`
```ocaml

  (** [print_rep repr] returns a string representation of [repr]. *)
  let print_repr r =
    let buf = Buffer.create 32 in
    let rec aux
      : type f. Buffer.t -> f fun_repr -> unit
      = fun buf r -> match r with
        | Fun_map (_, sub_repr) -> aux buf sub_repr
        | Fun_tbl r ->
           Buffer.add_string buf (Poly_tbl.print r.fun_tbl);  (* <-- here *)
           Printf.bprintf buf "_ -> %s" (match r.fun_arb.print with
              | None -> "<opaque>"
              | Some print -> print r.fun_default
            ); *)
    in
    Printf.bprintf buf "{";
    aux buf r;
    Printf.bprintf buf "}";
    Buffer.contents buf
```
But it's not a module that I use, don't really know what's the most suitable. It depends on if
`Poly_tbl.print` is mostly used throught `print_repr` or by itself.

vch9: I put the printer as an optional printer in those functions.

vch9: I did put optional printer at various places, see  https://github.com/c-cube/qcheck/pull/131/commits/05181aa1552aecb1f94af2800c5f382c9fbd7ceb.

vch9:
I also ""fixed"" unsigned int32 and int64 in https://github.com/c-cube/qcheck/pull/131/commits/154731b2306ff4474ca172df3496d4ce70f78170, with a very smart solution

vch9:
It is ready for a review @c-cube. The main is part is only removal of code and I believe the most questionable changes are on
the printer in `Poly_tbl` and `fun_ repr`.

vch9:
> ```
> # Int32.abs Int32.min_int;;
> - : int32 = -2147483648l
> ```
That's ugly, I was not aware of that behaviour

c-cube:
It's because min_int is `- 2^31`, whereas max_int is `2^31 - 1`,  so there's nowhere for `abs min_int` to go :slightly_smiling_face:

c-cube:
Alright, that looks good. Thank you!

PR #132, opened 2021-06-21
Fix signature of string

vch9:
The signature of `Gen.string` is a bit tricky to use.

```ocaml
open QCheck2
let foo : string Gen.t = Gen.string
```
is rejected because the last argument of `Gen.string` is a labelled argument.
You have to write:
```ocaml
open QCheck2
let foo : string Gen.t = Gen.string ?char:None
```
The solution I propose is to write:
```ocaml
open QCheck2
let foo : string Gen.t = Gen.string ()
```
This "problem" is introduced as we directly use `Gen` module instead of arbitrary.
Feel free to close the MR if you don't agree with that change.


c-cube:
I'd rather split into two different functions I think.

`string : string gen`, and `string_of: char gen -> string gen`, or something like that.

vch9:
> I'd rather split into two different functions I think.
> 
> `string : string gen`, and `string_of: char gen -> string gen`, or something like that.

`string_of` already exist, I've just changed the previous one for `string : string gen`

c-cube:
thank you!

Issue #135, opened 2021-06-24
Regressions for string printing/shrinking?


jmid:
First of all: thanks for an incredible effort on the integrated shrinking! 
I apologize for being late to the party - I finally found time to give QCheck2 a quick spin.
This is using  a fresh clone from master - loaded into utop with `#mod_use`.

* Testing a stupid property (all strings are non-empty) fails - but it is unclear that a counterexample is printed:
   ```
   # let t = Test.make ~print:Print.string Gen.string (fun s -> s <> "");;
   # QCheck_base_runner.run_tests ~verbose:true [t];;
   generated error fail pass / total     time test name
   [✗]    6    0    1    5 /  100     0.0s anon_test_5
   
   --- Failure --------------------------------------------------------------------
   
   Test anon_test_5 failed (0 shrink steps):
   
   
   ================================================================================
   failure (1 tests failed, 0 tests errored, ran 1 tests)
   ```
   
   Compared to QCheck1 this may be as simple as adding surrounding string quotes in the (empty) string printer:
   ```
   #open QCheck;;
   # let t = Test.make string (fun s -> s <> "");;
   # QCheck_base_runner.run_tests ~verbose:true [t];;
   random seed: 92576387
   generated error fail pass / total     time test name
   [✗]    1    0    1    0 /  100     0.0s anon_test_1
   
   --- Failure --------------------------------------------------------------------
   
   Test anon_test_1 failed (0 shrink steps):
   
   ""
   ================================================================================
   failure (1 tests failed, 0 tests errored, ran 1 tests)
  ```

* Testing an equally stupid string property (all strings equal their own concatenation) hits another issue:

   ```
   # let t = Test.make ~print:Print.string Gen.string (fun s -> s = s^s);;
   # QCheck_base_runner.run_tests ~verbose:true [t];;
   generated error fail pass / total     time test name
   [ ]    0    0    0    0 /  100     0.0s anon_test_6 (generating)Stack overflow during evaluation (looping recursion?).
   ```
   
   which didn't cause problems prior to integrated shrinking:

   ```
   # let t = Test.make string (fun s -> s = s^s);;
   # QCheck_base_runner.run_tests ~verbose:true [t];;
   generated error fail pass / total     time test name
   [✗]    1    0    1    0 /  100     0.0s anon_test_2
   
   --- Failure --------------------------------------------------------------------
   
   Test anon_test_2 failed (5 shrink steps):
   
   "\210"
   ================================================================================
[...]

PR #139, opened 2021-07-09
fix (print): Fix printing of strings and chars

Sir4ur0n:
Related to https://github.com/c-cube/qcheck/issues/135


c-cube:
simple, efficient :)

PR #143, opened 2021-07-13
fix (bytes): Fix bytes shrinking

Sir4ur0n:
* Fix bug in bytes/string shrinking (bad laziness + ref)
* Add test to prevent reintroducing this bug in the future

Closes https://github.com/c-cube/qcheck/issues/135

The added test fails (stackoverflow) on `master` but succeeds with the fix.


c-cube:
Awesome, especially the test!

PR #150, opened 2021-08-09
Lift 4.09 restriction on expect tests

jmid:
This is an attempt to lift the `< 4.09` requirement on @c-cube's expect tests.
It works by just omitting `file` from the `grep`ed error string. It has been verified on 4.11.2 and 4.12 locally,
I've furthermore tested the change by making a dummy modification in the `output.txt,expected`-files and verified that any such difference is caught and reported.

Motivation: I wanted to add some more (negative) expect tests of QCheck2(+1) and test that they run locally with 4.12.


c-cube:
awesome, thank you!

PR #154, opened 2021-08-15
Fix the tests and remove out of place (optional)

kit-ty-kate:
Fix all instance of `dune runtest -p qcheck*` encountered when testing through opam (see https://github.com/ocaml/opam-repository/pull/19207)


c-cube:
Merci beaucoup :grin:

Issue #155, opened 2021-08-16
`Gen.string_readable` can produce unreadable strings - misnamed?


jmid:
While testing `QCheck2` for PR #153 I realized that both `QCheck2.Gen.string_readable` and  `QCheck.Gen.string_readable` produce strings with the `Gen.char` generator - hence in the range 0-255. 

With the name `string_readable` as a first-time user, I had expected the generator to use `Gen.printable` as its underlying character generator - although I was also puzzled by the difference in the readable/printable terminology.

```ocaml
utop # QCheck.Gen.(generate1 string_readable);;
- : striÒ" = " ajRª;
utop # QCheck.Gen.(generate1 string_r\016yº\002~&dûWÛwí¤h>ÄÌI¯O6\022í4v"
utop # QCheck2.Gen.(generate1 string_readable);;
utop # QCheck2.Gen.(generate1 string_readable);;
- : string = "d¬C¤y3)\015"
utop # QCheck2.Gen.(generate1 string_readable);;
- : string = "¾\015ç"
```
Above, the two samples from `QCheck.Gen.string_readable` and the first sample from `QCheck2.Gen.string_readable` produced strings with suitable control characters (backspace, clear?) to mess up the `utop` output - bonus points to `QCheck2` for deleting the sample entirely!

jmid:
BTW: `Gen.string_readable` was added to address issue #75.

c-cube:
My bad, I was the origin of the bad name. I just fixed it by deprecating the `string_readable` in favor of the new `string_printable`, and both use `Gen.printable` to pick chars.

jmid:
I think there are two orthogonal issues here: (a) the generator's name and (b) the generator's functionality. We agree there was a mismatch between the two.

The proposed fix in 727dcc5 
- adds a new generator `Gen.string_printable` where the two are better aligned  😀
- deprecates `Gen.string_readable` 

`Gen.string_readable` was added to address issue #75 as an alias of `Gen.string` but without an optional parameter (which was causing problems type-wise). As far as I can tell `Gen.string_printable` does not fill the hole that  `Gen.string_readable`'s functionality was intended to fill ("string generator w/char codes in 0-255 like `Gen.string`, but without an optional parameter"):
```ocaml
  let string_readable = string_printable
```
[...]

PR #158, opened 2021-08-18
QCheck2.Gen.string_readable -> QCheck2.Gen.string_printable

jmid:
Following up on issue #155, this PR
- renames `QCheck2.Gen.string_readable` to `string_printable`
- updates its `char` generator to use `Gen.printable`
- updates the documentation
- updates the tests

As far as I can tell, QCheck2 hasn't been released yet, so we can do a hard renaming (and avoid a deprecation).

CC @sir4ur0n @vch9: are you OK with this change?


jmid:
Ha! I (naturally) forgot to update the expected output - but the CI tests caught it. 😄

sir4ur0n (approved):

sir4ur0n:
> CC @sir4ur0n: are you OK with this change?

Of course :smile: This is a good change, thank you!

vch9:
I think the hard renaming is ok, however you could add it to the changelog.

vch9:
> CC @vch9: are you OK with this change?
Absolutely, thank you! :smiley:

Issue #156, opened 2021-08-17
QCheck2.Gen.list_size stackoverflow on big sizes


sir4ur0n:
**`QCheck2.Gen.list_size` crashes with `Stack_overflow`**

I added a few tests with a sized list generator from issue #64 which crashes the `QCheck2.Gen.list_size` generator with `Stack_overflow`. I believe that this is because it isn't tail recursive like `foldn` underlying `QCheck.Gen.list_size`:
https://github.com/c-cube/qcheck/pull/153/files#diff-44df483bf5666cb8dd729f02820e34cec6c7b7199bedcb551d82fcfccdff8c05R160-R176
Can it be formulated with an accumulator instead?

_Originally posted by @jmid in https://github.com/c-cube/qcheck/issues/153#issuecomment-898711290_

sir4ur0n:
I took a look at it yesterday and today.

:warning: Disclaimer: I am not seasoned in performance/debugging in OCaml, so my investigations should be taken with a grain of salt. Any more experienced person is more than welcome to correct me or help me better analyze the problem :grin: 

Making `list_size` tail recursive is apparently not enough: I still had `StackOverflow` errors (but to my surprise, the error backtrace only contained 4 stack layers...):
```ocaml
Test lists equal to duplication failed:

ERROR: uncaught exception in generator for test lists equal to duplication after 100 steps:
Exception: Stack overflow
Backtrace: Raised by primitive operation at QCheck2.Gen.liftA2 in file "src/core/QCheck2.ml", line 247, characters 4-18
Called from QCheck2.Gen.ap in file "src/core/QCheck2.ml", line 239, characters 74-80
Called from QCheck2.Tree.bind in file "src/core/QCheck2.ml", line 191, characters 28-31
Called from QCheck2.Test.check_state in file "src/core/QCheck2.ml", line 1520, characters 12-32
```

Note that I added `let () = Printexc.record_backtrace true` before the test to get backtraces.

And this is weird to me, cause I am used in other languages to having a HUGE stack printed whenever a StackOverflow happens. I have no idea why in OCaml this happens...

That being said, I noticed that replacing `list_size` implementation (even tail recursive) with a `for` loop seems to work?

```ocaml
(* Current implementation on master.
   The stackoverflow contains a huge amount of recursive calls of QCheck2.Gen.ap which is called by liftA2 *)
  let list_size (size : int t) (gen : 'a t) : 'a list t =
    size >>= fun size ->
    let rec loop n =
[...]

PR #160, opened 2021-08-18
Fix `QCheck2.gen.list_size Stack overflow`

jmid:
This PR addresses issue #156. It 
- replaces `QCheck2.gen.list_size` with a tail-recursive generator over `Tree.t` similar to @sir4ur0n's suggestion
- updates the expected test outputs
- removes the custom length printer from the `list_equal_dupl` test (to ensure that both counterexamples are `[0]`)

### DIFF
--- a/test/core/QCheck2_expect_test.ml
+++ b/test/core/QCheck2_expect_test.ml
@@ -143,7 +143,7 @@ module Shrink = struct
 
   let strings_are_empty =
     Test.make ~name:"strings are empty" ~count:1000 ~print:Print.string
-      Gen.string (fun s -> (*Printf.printf "\"%s\"\n" (String.escaped s);*) s = "")
+      Gen.string (fun s -> s = "")
 
   let string_never_has_000_char =
     Test.make ~name:"string never has a \\000 char" ~count:1000 ~print:Print.string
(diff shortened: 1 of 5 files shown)


test/core/QCheck2_expect_test.ml
```diff
@@ -182,7 +182,7 @@ module Shrink = struct
       (fun xs -> (*print_list xs;*) List.length xs < 4332)
 
   let list_equal_dupl =
-    Test.make ~name:"lists equal to duplication" ~print:length_printer
+    Test.make ~name:"lists equal to duplication" ~print:Print.(list int)
       Gen.(list_size size_gen small_int) (*Gen.(list small_int)*)
```
sir4ur0n, line 186: Why keep some commented out code, like this, or like in `test/core/QCheck_expect_test.ml`?

sir4ur0n (approved):
Looks good (besides the minor comment I added), thank you

test/core/QCheck2_expect_test.ml
jmid: These were old combinations that I had considered.
Leaving them as comments was not ... ideal 😄
They have been removed in the latest commit.

PR #161, opened 2021-08-18
Fix `QCheck2.Gen.small_string` impl to match interface and documentation

jmid:
This PR updates the `QCheck2.Gen.small_string` to match the interface and documentation following my finding in #153:
```ocaml
      Accepts an optional character generator (the default is {!char}).

      Shrinks on the number of characters first, then on the characters.
  *)
```

### DIFF
--- a/src/core/QCheck2.ml
+++ b/src/core/QCheck2.ml
@@ -661,7 +661,7 @@ module Gen = struct
 
   let string_readable = string_size ~gen:char nat
 
-  let small_string ?gen st = string_size ?gen small_nat st
+  let small_string ?(gen = char) st = string_size ~gen small_nat st
 
   let small_list gen = list_size small_nat gen
 


vch9:
:+1:

jmid:
It is nice to align the `.ml` and `.mli` for consistency.
However, as far as I can tell (see examples and discussion in #162) this still means that the parameter is defacto labelled - and thus required. Merging this fix does not change that - meaning the documentation+interface should be updated 😬🤷‍♂️
As such I'm hesitant with the merge, because we will have to make some additional changes anyway...

I'd therefore like your input on the discussion in #162 @vch9 @sir4ur0n @c-cube 
(I'd also be grateful if one of you would confirm by cloning, running `dune utop` and pasting the examples)

sir4ur0n (approved):

sir4ur0n (approved):
While it doesn't solve the "optional mandatory" argument design issue you mentioned, it doesn't hurt in any way, so it's still positive :+1:

jmid:
Closing as this is addressed in #245

Issue #164, opened 2021-08-18
Different `Gen.int` distributions


jmid:
The regression `diff -y` finds a difference in average, standard deviation, etc. of `Gen.int` whereas the remaining generators reveal the exact same distribution:
```
+++ Stats for int dist ++++++++++++++++++++++++++++++++++++++++++++++++++++++	+++ Stats for int dist ++++++++++++++++++++++++++++++++++++++++++++++++++++++

stats dist:									stats dist:
  num: 100000, avg: 2541076923587387.50, stddev: 2660730801206827008.00, medi |	  num: 100000, avg: 6126662802445055.00, stddev: 2661484817981980672.00, medi
  -4611522359435274428..-4150369195341695293: ############################### |	  -4611578348806740501..-4150415539984062486: ###############################
  -4150369195341695292..-3689216031248116157: ############################### |	  -4150415539984062485..-3689252731161384470: ###############################
  -3689216031248116156..-3228062867154537021: ############################### |	  -3689252731161384469..-3228089922338706454: ###############################
  -3228062867154537020..-2766909703060957885: ############################### |	  -3228089922338706453..-2766927113516028438: ###############################
  -2766909703060957884..-2305756538967378749: ############################### |	  -2766927113516028437..-2305764304693350422: ###############################
  -2305756538967378748..-1844603374873799613: ############################### |	  -2305764304693350421..-1844601495870672406: ###############################
  -1844603374873799612..-1383450210780220477: ############################### |	  -1844601495870672405..-1383438687047994390: ###############################
  -1383450210780220476.. -922297046686641341: ############################### |	  -1383438687047994389.. -922275878225316374: ###############################
   -922297046686641340.. -461143882593062205: ############################### |	   -922275878225316373.. -461113069402638358: ###############################
   -461143882593062204..       9281500516931: ############################### |	   -461113069402638357..      49739420039658: ###############################
         9281500516932..  461162445594096067: ############################### |	        49739420039659..  461212548242717674: ###############################
    461162445594096068..  922315609687675203: ############################### |	    461212548242717675..  922375357065395690: ###############################
    922315609687675204.. 1383468773781254339: ############################### |	    922375357065395691.. 1383538165888073706: ###############################
   1383468773781254340.. 1844621937874833475: ############################### |	   1383538165888073707.. 1844700974710751722: ###############################
   1844621937874833476.. 2305775101968412611: ############################### |	   1844700974710751723.. 2305863783533429738: ###############################
[...]

PR #165, opened 2021-08-18
Align int generators

jmid:
The PR addresses issue #164 from PR #153.

- It fixes the generation order of both QCheck/QCheck2 `int` generators to be left-middle-right like `QCheck.Gen.int`. This should be preferable to continue past behaviour and for reproducability.
- It updates the expected QCheck2 output (bonus: with this change all reported stats are *identical*, yay!)
- It pull out a `Sys.word_size` test, so that `QCheck2.Gen.pint_raw` only tests it once (not on every generator call)


jmid:
OK, I´m merging this one since it shouldnt be too controversial.

Issue #127, opened 2021-06-11
ppx_deriving_qcheck


vch9:
Hello !

I just want to let you know that I'm currently working on a ppx for qcheck.
My main focus is to produce a `t arbitrary` for (any?) `t`. 

Here's the project:
https://github.com/vch9/ppx_deriving_qcheck

There is no release yet, but feel free to take a look if you're curious.
Especially in the tests, you'll be able to see what the generated code
looks like.

c-cube:
Awesome! I'll take a look :)
I've thought about something like that for a long time but never followed through.

c-cube:
I think it'd be nice to name it `[@@deriving qcheck]`, no? it'd be more explicit…

vch9:
`[@@deriving arb]` might no be the most explicit name, but wouldn't `[@@deriving qcheck]` be to general ?

c-cube:
too general in what sense? I don't see with what it could be confused.

vch9:
I don't really know the conventions about a deriving name. It's true that the ppx, I believe, will not derive anything else ?
That might be too much but we could maybe have 2 rules:
```
type t = A | B | C [@@deriving qcheck]
```
who would generate a `QCheck.arbitrary`

and
```
type t = A | B | C [@@deriving generator]
```
who would create a `unit -> t` generator, abstracting the QCheck library.

vch9:
How would you feel about adding ppx_deriving_qcheck into qcheck ? I'm currently cleaning and improving the code right now.

This would avoid to avoid a dependency from qcheck **and** ppx_deriving_qcheck.  Beside the advantages of a deriver, the "exhibition" given to the deriver would avoid people to work on a similar tool.

c-cube:
> How would you feel about adding ppx_deriving_qcheck into qcheck? I'm currently cleaning and improving the code right now.

that sounds like a good idea.

> This would avoid to avoid a dependency from qcheck **and** ppx_deriving_qcheck.

I don't see how, a ppx deriver would still be its own package. It makes sense for it to live in the same repo, but not to be the same package.

vch9:
> I don't see how, a ppx deriver would still be its own package. It makes sense for it to live in the same repo, but not to be the same package.

Yep right my bad, we would still have both packages.

Well, I'll be soon creating a MR about the package.

PR #171, opened 2021-09-04
ppx_deriving_qcheck

vch9:
Well as discussed in #127 , here it is.

As far as I known, "only" GADT are not supported. Feel free to give me types we should
add to the test suite.

If you want examples of derivation, `test/ppx_deriving_qcheck/deriver/test.ml` is a good starting point. If any types
in expected fields are wrong, well, the tool is wrong.

Merging the deriver is nothing urgent, but at least the discussion will be centralized in this main repo. This will avoid
people writing another deriver doing the exact same things.

On a side node, I have added my self to the Authors so I can be reachable If the deriver needs to be updated.


vch9:
I have some adjustments to do for making my deriver compiles on 4.08

c-cube:
@jmid you think this should be merged as is?

jmid:
> @jmid you think this should be merged as is?

I think this is a great effort and contribution. Well done @vch9!
I was particularly happy to see a bunch of tests included 😀

A few remarks:
- I agree that including the deriver in the qcheck repo would bring visibility and thus avoid duplicate efforts
- I think the sub-library should be highlighted in the main README.md, like qcheck-alcotest
  The sub-library's own README.md should probably be expanded a bit, e.g., with a complete example and its output, listing both the supported types and the unsupported ones
- I don't think we should make the deriver a mandatory part of the `qcheck` package. This would mean existing qcheck users, users of QCheck2, and CI builds, would all now require `ppxlib` without needing it. Here I would rather favor a "pay-as-you-go" approach.

I wonder how the deriver works for mutually recursive data types which reference each other? Consider:
```ocaml
  type stmt =
    | Assign of string * aexp
    | Block of block
    | If of relexp * stmt
    | While of relexp * stmt
  and block = Bl of stmt list
```
adapted from https://github.com/jmid/sm2-tes21/blob/main/lec08/stmtlang.ml

Here I would hand-write two sized generators that pass the fuel parameter between each other
to avoid that each one generates a large new "fuel amount" and thus letting the generation
continue deeply.
I do so in https://github.com/jmid/sm2-tes21/blob/5c1d9406b8457560999cf031c0c808011590bf70/lec08/stmtlang.ml#L94-L110 
(but over a pair of `stmt` and `stms list` generators rather that with an explicit `block`-type like above):
```ocaml
  let rec stmtgen env = fix (fun rgen n -> match n with
      | 0 -> assign_gen env n
      | _ ->
        oneof
          [assign_gen env n;
           map  (fun ss   -> Block ss)     (stmtlistgen env (n-1));
           map2 (fun re s -> If (re,s))    (relexpgen env (n/2)) (rgen (n/2));
           map2 (fun re s -> While (re,s)) (relexpgen env (n/2)) (rgen (n/2));
          ])
  and stmtlistgen env n = match n with
    | 0 -> return []
    | _ ->
      stmtgen env (n/2) >>= fun s ->
          let env' = (match s with
            | Assign (x,_) -> if List.mem x env then env else x::env
            | _ -> env) in
          stmtlistgen env' (n/2) >>= fun ss -> return (s::ss)
```


Finally: What's up with the separate license at the top of deriver/test.ml?

c-cube:
>     * I don't think we should make the deriver a mandatory part of the `qcheck` package. This would mean existing qcheck users, users of QCheck2, and CI builds, would all now require `ppxlib` without needing it. Here I would rather favor a "pay-as-you-go" approach.

100% agree. The `qcheck` library includes the other ones purely for backward compatibility (from a few years ago). The ppx shouldn't be part of it!

vch9:
> > ```
> > * I don't think we should make the deriver a mandatory part of the `qcheck` package. This would mean existing qcheck users, users of QCheck2, and CI builds, would all now require `ppxlib` without needing it. Here I would rather favor a "pay-as-you-go" approach.
> > ```
> 
> 100% agree. The `qcheck` library includes the other ones purely for backward compatibility (from a few years ago). The ppx shouldn't be part of it!

I am removing the deriver from libraries inside `src/dune` file. However, `opam install . --deps-only` will still install ppxlib as it is mentioned in the `deriver.opam`

vch9:
> Finally: What's up with the separate license at the top of deriver/test.ml?

Removed this (57cc95e3d2232097443fbf75be33fca87cc72808), I worked with that license when the project was outside that repo.

vch9:
> The sub-library's own README.md should probably be expanded a bit, e.g., with a complete example and its output, listing both the supported types and the unsupported ones

Will do.

vch9:
> I think the sub-library should be highlighted in the main README.md, like qcheck-alcotest

Done [here](6bb586f39c4e90e1294285af3261c8e6914aa4ec)

vch9:
>     1. Could you expand the first example to an example of a complete test?
>     This is with the best of intensions: A new-comer would benefit from a complete example (datatype, `Test.t`, + running it with a runner or `Test.check_exn`) before starting to read about the supported/unsupported types.

I've added in the README an example of test alongside the ppx annotation: 58cd4bc8ef184f21fcbd4d03aab54ec9a6a135bd.

>     2. Regarding the mutually recursive datatypes: This limitation should probably be addressed (either now or in a separate issue). I think it could be solved like the hand-written example, by deriving a collection of fueled/sized generators for each datatype (`gen_stmt_sized`,`gen_block_sized`), letting these be the interface that drives mutually-recursive generators, and finally have a little top-level wrapper that passed them each to `sized`.

Yes, this will be addressed. They are few small issues I need to work on (including this one, the biggest one). A special label `deriver` could be useful to distinguish issues in this repo. This would make things easier for me, I could solve one issue at the time during my free time.

Once you believe we can merge this @jmid, I'll squash fixup commits

vch9:
The MR is ready to be merged if you're ok with the current state @jmid. 
I'll be opening multiple issues following the merge.

PR #174, opened 2021-09-08
Test.check_exn and more unit tests

jmid:
When writing the shrinking logs I noticed that the `Test.check_exn` was wrong:
It said it would throw a `Test_error` exception on both failures and errors.
I figured it would be nice to add a couple of unit-tests to ensure this for `QCheck2`.

This PR does both of these.

As a bonus, it adds a few unit tests for `QCheck`, including some for the `int*` shrinkers
to bring the unit test status closer to `QCheck2`.

After having realized that the shrinker's output depend on whether the tested property holds or not,
I split them up into a successful shrink trace and a failure shrink trace,
I then noticed that our current `int*` shrinkers produce the same number twice at the end, e.g.,
for `Shrink.int 100` tested on, e.g., `(fun i -> i<100)` it tries: `[50; 75; 88; 94; 97; 99; 99]`  🤔
This is confirmed by `shrink_algo_logs/int_smaller_209609_qcheck.expected` which tests 209608 twice at the end.

I'll investigate. I think that could make for another potential `Shrink.int` improvement on top of PR #173.


jmid:
BTW: This PR builds on top of #172 - which renamed the QCheck2 unit test file.

jmid:
Quick update on the `int` shrinking:
https://github.com/c-cube/qcheck/blob/189a0258c96e44a2af0eceadc7e8aa669af8bc75/src/core/QCheck.ml#L470-L477
When `x` is 3, say, we will enter the loop, setting `y := 1` and yielding `2`.
After resuming we will then exit the loop only to yield `3-1 = 2` again.

PR #173 doesn't seem to have this problem.

jmid:
Superseeded by #233 and #234

Issue #108, opened 2021-03-16
Provide a more generic combinator (similar to Crowbar `gens`)


Sir4ur0n:
(This should be my last issue in the foreseeable future :grin: )

Currently QCheck provides `pair`, `triple` and `quad` to combine arbitraries into tuples, and then `Test.make` takes a **single** arbitrary to write the test.

So if one wants 3 ints, 3 strings and 3 bools, one must write one of the following:
```ocaml
Test.make (pair (pair (quad int int int string) (quad string string bool bool)) bool) (fun (((i1, i2, i3, s1), (s2, s3, b1, b2), b3) -> ...)

Test.make (triple (triple (int int int) (triple string string string) (triple bool bool bool)) (fun ((i1, i2, i3), (s1, s2, s3), (b1, b2, b3) -> ...)

(* etc. *)
```

(Yes, 9 arguments is a bit extreme, but in my experience 5 or 6 happens, and the problems below occur even with a single `pair`)

This has several quality-of-life problems:
* whenever a param needs to be added or removed, it is cumbersome to adapt the pairs/triples/quadruples in both the generator combination, and in the function arguments
* it "forces" arbitrary (haha) grouping of arbitraries/arguments together
* No currying (which also induces "no partial application")
* This is verbose

Note that none of the problems above is a show stopper. This is purely about improving the dev experience.

One thing I really like in Crowbar is the flat list of generators, which solves both problems above, thanks to [the `gens` type](https://github.com/stedolan/crowbar/blob/master/src/crowbar.ml#L27) (recursive type wrapping together the generator of `'a` and the function that consumes an `'a`) and [`add_test`](https://github.com/stedolan/crowbar/blob/master/src/crowbar.mli#L200).

With Crowbar the example above is written:
```ocaml
add_test [int; int; int; string; string; string; bool; bool; bool] (fun i1 i2 i3 s1 s2 s3 b1 b2 b3 -> ...)
```

Would such a thing be possible in QCheck?

If not, at least could we consider providing combinators up to 8, with less verbose names? E.g. `t2/t3/t4/t5/t6/t7/t8` rather than `pair/triple/quad/???`. While this would not completely solve all the problems listed above, this would decrease verbosity.

Note: I saw the [`Tuple` submodule](https://c-cube.github.io/qcheck/0.17/qcheck-core/QCheck/Tuple/index.html) but the only documentation that mentions it is about arbitrary functions, I'm not even sure this is usable for the need of this issue

[...]

PR #177, opened 2021-09-13
Shrinker performance benchmark

jmid:
This PR adds a shrinker benchmark.
It iterates the (factored) shrinker tests in QCheck_tests.ml and QCheck2_tests.ml in pairs.
For each such pair of tests we measure
- time in seconds, denoted `Q1/s` and `Q2/s`
- the number of succesful shrinks, denoted `#succ`
- the number of shrink attempts, denoted `#att`

This is iterated across several seeds to make sure observations hold across more than one random run.
Finally we add up the timings to spot general trends.

Overall, the benchmark gives a reasonable idea of the state of affairs (see output below):
- the QCheck list shrinker tries *way* too many candidates, often making it unbearably inefficient on large lists, e.g, spending 9.8 secs over 5 milion shrink attempts of which only 1.696 are successful(!)
- the QCheck string shrinker does not perform as well as the QCheck2 shrinker, e.g., spending 1.342 seconds on 4465 shrink-steps when QCheck2 gets away with only 24 in 0.002 secs
- the QCheck2 list shrinker could also use a hand, e.g., spending 17.010 secs (twice of QCheck) on `lists shorter than 4332`
- the function shrinkers in both QCheck and QCheck2 suffer from poor list-shrinking and the involved domain+range shrinkers, e.g., QCheck2 spending 15.278 secs on `fold_left fold_right uncurried` in one iteration, whereas QCheck spends  28.631 secs on only 125 successful shrink attempts out of 27750 on `fold_left test, fun first` in another iteration.

The benchmark starts to drill the framework:
- On one seed I had the segfault reported in #175 
- I've also had to cherry-pick random seeds that don't make the benchmark take hours on my laptop.

I've therefore drafted (QCheck) performance improvements that will make such hacks unnecessary. I'll commit them separately.

```
                                                         iteration seed 1234                   iteration seed 8743                   iteration seed 6789               total
Shrink test name                                  Q1/s  #succ/#att   Q2/s  #succ/#att   Q1/s  #succ/#att   Q2/s  #succ/#att   Q1/s  #succ/#att   Q2/s  #succ/#att    Q1/s   Q2/s
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
big bound issue59                                - skipped as generator is stateful, making it non-repeatable
long_shrink                                       0.051  149/351     1.700 3039/3099    0.010  148/349     1.064 3068/3127    0.013  146/345     1.172 3063/3124    0.074  3.937
[...]

PR #181, opened 2021-09-15
Add tup2 to tup9 for Gen

vch9:
Alternative (and easy) solution for #108.

Nothing smart happening there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changes
 
+## 0.19
+
+- add tup2 to tup9 for generators
+
 ## 0.18
 
 This releases marks the addition of `QCheck2`, a module where generation
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -42,6 +42,28 @@ let _opt_map_4 ~f a b c d = match a, b, c, d with
   | Some x, Some y, Some z, Some w -> Some (f x y z w)
   | _ -> None
 
+let _opt_map_5 ~f a b c d e = match a, b, c, d, e with
+  | Some x, Some y, Some z, Some u, Some v -> Some (f x y z u v)
+  | _ -> None
+
+let _opt_map_6 ~f a b c d e g = match a, b, c, d, e, g with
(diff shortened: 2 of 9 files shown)


jmid:
I think these are fine. What do you think @c-cube?

Even if they are so straightforward, it would be nice to include tests of them.
Doing so, will require `Print`ers - which should be included for the generators to be generally useful.

For uniformity it would be nice to add `t2`-`t9` to `QCheck` as well with both generators and printers - and ideally also shrinkers...

c-cube:
can we get the same in QCheck.Gen?

vch9:
Yes I will add these in QCheck. 

When I did this PR I thought QCheck2 would replace QCheck. However, I started the migration in our project where we have a _lot_ of QCheck tests and the execution's time has raised a lot. I now believe QCheck is still relevant in many cases.

c-cube:
Indeed, it might be that they both have their use :-).

vch9:
It is ready again for a review @c-cube and @jmid. I added printers, shrinkers, generators and arbitraries? for QCheck and printer for QCheck2. I also added a lot of tests aswell. It's a lot of boiler-plate code thought.

c-cube:
Looks good I think. Thanks for the patience on such boilerplat-y code :D

jmid:
>  However, I started the migration in our project where we have a _lot_ of QCheck tests and the execution's time has raised a lot. 

It would be great if you could extract a representable test case and share it here 🙏‍🙂

QCheck2 is a great effort! In the long run, I still hope for it to replace QCheck.
It will require some more engineering, testing, and fine-tuning evidently.
A real-world example test that takes too long w/QCheck2 would be a good starting point though (in comparison, the shrinker benchmarking in #177 is mostly based on artificial tests).

vch9:
> A real-world example test that takes too long w/QCheck2 would be a good starting point though (in comparison, the shrinker benchmarking in #177 is mostly based on artificial tests).

It's a test in the Tezos code base, there is a lot of dependencies going on there. I don't know if it can be easily extracted for our testing purpose.

Issue #180, opened 2021-09-14
`nat_split2` and` pos_split2` are reversed


just-max:
(documentation snippets are from https://c-cube.github.io/qcheck/0.18/qcheck-core/QCheck/Gen/index.html)
Expected behaviour: generators with "nat" in their names should generate non-negative natural numbers, generators with "pos" should generate strictly positive natural numbers.

The documentation agrees:
- `val nat_split2 : int -> (int * int) t` : ... generates pairs (n1, n2) of natural numbers ...
- `val pos_split2 : int -> (int * int) t` : ... generates pairs (n1, n2) of strictly positive (nonzero) natural numbers ...

What actually happens:
- nat_split2 cannot generate zeroes, even though it should (ie. `QCheck.Gen.(generate (nat_split2 2))` only generates pairs `(1, 1)`)
- pos_split2 can and does generate zeroes, even though it shouldn't (ie. `QCheck.Gen.(generate (pos_split2 0))` does not cause errors and generates pairs `(0, 0)`)
- The documentation for `nat_split2` also states the restriction "raises Invalid_argument unless n >= 2" which should really apply to `pos_split2`.
- The documentation for `pos_split2` refers to `nat_split2` (... "`nat_split2 n` generates pairs")

Although it should probably be a separate issue, the counterparts generalized to lists, `nat_split` and` pos_split`, don't seem to work at all (regardless if the precondition `n <= k` is satisfied).

gasche:
I am probably responsible for the mistakes (I proposed `nat_split` and `pos_split` in #114), I'll have a look.

PR #176, opened 2021-09-11
More tests

jmid:
This PR 
- adds tuple and bind tests (both positive, negative, and statistics).
- I also added a manual shrinker and test for the `IntTree` test.
- I grouped test-names per sub-module as the long list at the end was getting hard to maintain.

*Edit:* I've now also
- added shrink-logging tests to compare the total number of shrinking attempts.
- separated the tests into a separate "test library" and runner.
Then we have one runner for the expect tests in the CI - while other runners can reuse (some of) the same tests,
e.g., for local shrinker benchmarking.

Here's the `diff -y` of the new test outputs. Note how QCheck2's int shrinking strategy generally spends
less successful shrinking steps (again) - as have been discussed in, e.g., PR #153 and #173:
```
--- Failure --------------------------------------------------------	--- Failure --------------------------------------------------------

Test pairs have different components failed (0 shrink steps):		Test pairs have different components failed (0 shrink steps):

(4, 4)									(4, 4)

--- Failure --------------------------------------------------------	--- Failure --------------------------------------------------------

Test pairs have same components failed (125 shrink steps):	     |	Test pairs have same components failed (63 shrink steps):

(0, 1)									(0, 1)

--- Failure --------------------------------------------------------	--- Failure --------------------------------------------------------

Test pairs have a zero component failed (124 shrink steps):	     |	Test pairs have a zero component failed (122 shrink steps):

(-1, 1)								     |	(1, 1)

--- Failure --------------------------------------------------------	--- Failure --------------------------------------------------------

Test pairs are (0,0) failed (125 shrink steps):			     |	Test pairs are (0,0) failed (63 shrink steps):

(0, 1)									(0, 1)

--- Failure --------------------------------------------------------	--- Failure --------------------------------------------------------

Test pairs are ordered failed (125 shrink steps):		     |	Test pairs are ordered failed (2 shrink steps):

(0, -1)									(0, -1)

--- Failure --------------------------------------------------------	--- Failure --------------------------------------------------------

Test pairs are ordered reversely failed (125 shrink steps):	     |	Test pairs are ordered reversely failed (63 shrink steps):

(0, 1)									(0, 1)

--- Failure --------------------------------------------------------	--- Failure --------------------------------------------------------

Test pairs sum to less than 128 failed (121 shrink steps):	     |	Test pairs sum to less than 128 failed (59 shrink steps):

(0, 128)								(0, 128)

--- Failure --------------------------------------------------------	--- Failure --------------------------------------------------------

[...]

PR #183, opened 2021-09-22
fix Gen.{nat,pos}_split{2,}

gasche:
This is a fix for several bugs reported in #180.

Those functions (and `range_subset`), implemented by myself in #114, were utter garbage. I had tested their implementation heavily within a different random library, and when I ported them to QCheck it seems that I messed things up considerably (this is fiddly code full of off-by-one traps etc.).

In the present PR I have gone over the implementation of each function and done some manual testing, and I also wrote automated tests in the existing QCheck testsuite. (This forced me to think about the valid input space for the combinators; I extended `pos_split` to support the case `size=0, n=0`, and found a treacherous bug when `size=1`.)

My recommendation for the future would be to encourage new generators to come with their tests.

Sorry for the mess, and thanks @max-lang for the report!

### DIFF
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -248,27 +248,27 @@ module Gen = struct
     List.sort (fun (w1, _) (w2, _) -> poly_compare w1 w2) samples |> List.rev_map snd
 
   let range_subset ~size low high st =
-    if not (low <= high && size <= high - low + 1) then invalid_arg "Gen.range_subset";
+    let range_size = high - low + 1 in
+    if not (0 <= size && size <= range_size) then
+      invalid_arg "Gen.range_subset";
     (* The algorithm below is attributed to Floyd, see for example
        https://eyalsch.wordpress.com/2010/04/01/random-sample/
        https://math.stackexchange.com/questions/178690
 
-       Note: the code be made faster by checking membership in [arr]
-       directly instead of using an additional Set. None of our
-       dependencies implements dichotomic search, so using Set is
-       easier.
+       Note: the code is easier to read when drawing from [0..range_size-1]
+       rather than [low..high]. We draw in [0..bound], and shift the
+       results by adding [low] when writing them to the result array.
     *)
     let module ISet = Set.Make(Int) in
     let s = ref ISet.empty in
-    let arr = Array.make size 0 in
-    for i = high - size to high do
-      let pos = int_range high i st in
-      let choice =
-        if ISet.mem pos !s then i else pos
-      in
-      arr.(i - low) <- choice;
+    for i = range_size - size to range_size - 1 do
+      let pos = int_range 0 i st in
+      let choice = if ISet.mem pos !s then i else pos in
       s := ISet.add choice !s;
     done;
+    let arr = Array.make size 0 in
+    let idx = ref 0 in
+    ISet.iter (fun choice -> arr.(!idx) <- low + choice; incr idx) !s;
     arr
 
   let array_subset size arr st =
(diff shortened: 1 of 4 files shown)


src/core/QCheck.ml
```diff
@@ -334,31 +334,36 @@ module Gen = struct
...
 
-  let nat_split2 n st =
-    if (n < 2) then invalid_arg "nat_split2";
+  let pos_split2 n st =
+    if (n < 2) then invalid_arg "pos_split2";
     let n1 = int_range 1 (n - 1) st in
     (n1, n - n1)
 
-  let pos_split2 n st =
+  let nat_split2 n st =
```
just-max, line 342: Looks good, thanks for fixing this.
Should `nat_split2` raise an exception too, for consistency?
```suggestion
  let nat_split2 n st =
  if (n < 0) then invalid_arg "nat_split2";
```
Sure, `int_range` will complain anyway, but this makes the source of the error clearer.

gasche: My approach so far has been to skip the "obvious" tests that all parameters should be non-negative, and test the "interesting" relations (strictly positive bounds, or order relations between the parameters). Your suggestion does not hurt either, and it sort of reminded me that I didn't include tests for `range_subset`, so I'll add that to the PR in any case.

gasche: I made the two changes. Thanks!

vch9:
Maybe we could add these generators in `QCheck2` as well?

gasche:
I don't know what the current status is for QCheck vs. QCheck2, is there some explanation somewhere?

vch9:
> I don't know what the current status is for QCheck vs. QCheck2, is there some explanation somewhere?

I believe we kept `QCheck` for retro-compatibility issues (ping @c-cube, @sir4ur0n). Tests from `"QCheck1"` are actually migrated to `QCheck2`: https://github.com/c-cube/qcheck/blob/b065a81661d312fc7bbdbf95fb24229d71991a10/src/core/QCheck2.mli#L1753.

Because `QCheck2` is new (released ~last week), we can afford breaking changes; such as designs changes, remove/rename generators.

gasche:
I think it would be nice to (review and) merge this PR without waiting too much, because it fixes real bugs in released code. I don't mind duplicating the logic in QCheck2.ml, but I think this is a somewhat independent question that can be handled in another PR.

In any case, I wish the current status of core/QCheck.ml vs. core/QCheck2.ml was clearly stated somewhere, for example in the README, so that other occasional contributors know what to do.

vch9:
> I think it would be nice to (review and) merge this PR without waiting too much, because it fixes real bugs in released code. I don't mind duplicating the logic in QCheck2.ml, but I think this is a somewhat independent question that can be handled in another PR.

yes

> In any case, I wish the current status of core/QCheck.ml vs. core/QCheck2.ml was clearly stated somewhere, for example in the README, so that other occasional contributors know what to do.

and yes

vch9:
LGTM, tested bugs reported in #180 are fixed.

c-cube:
> I don't know what the current status is for QCheck vs. QCheck2, is there some explanation somewhere?

That's a good question actually. I think both still have their place, even if only because `QCheck` is a bit simpler conceptually. Maybe in the future we'll actually deprecate `QCheck` but I think it'll take years, and will require good confidence that the shrinking of `>>=` and others is good. For now, in my mind, both are supported.

test/core/QCheck_expect_test.ml
```diff
@@ -127,6 +136,70 @@ module Generator = struct
...
+           && Array.for_all (fun (a, b) -> a < b)
+               (Array.init (size - 1) (fun k -> arr.(k), arr.(k+1)))
+           && arr.(size - 1) <= high)
+
+  let nat_split =
+    Test.make ~name:"nat_split"
+      (make
+         ~print:Print.(pair int (array int))
+         Gen.(small_nat >>= fun n ->
+              pair (return n) (nat_split ~size:n n)))
```
jmid, line 185: #198 reminded me of this PR - so I was just looking at the code.

If I understand correctly, this tests an `n`-way split of `n`, which is an important case to test! 
To also test the case `size < n` can some `min/max` magic like for `pos_split` (below) be applied?

test/core/QCheck_expect_test.ml
```diff
@@ -127,6 +136,70 @@ module Generator = struct
...
+      (make
+        ~print:Print.(pair int (pair int int))
+         Gen.(small_nat >>= fun n ->
+              (* we need n > 2 *)
+              let n = n + 2 in
+              pair (return n) (pos_split2 n)))
+      (fun (n, (a, b)) ->
+         (0 < a && 0 < b && a + b = n))
+
+  let range_subset_spec =
```
jmid, line 160: Since `range_subset` comes with a "distribution spec", would it make sense to supplement with a stat expect test? "Many repetitions of `~size:1` distribute evenly in a histogram", "for a single run with large `size` and larger range the avg/median/variance come out as expected", ...? (just a thought)

gasche: I want to avoid probabilistic tests that fail with non-zero probability.

(Note: I wrote a "reference implementation" of these functions [in random-generator](https://github.com/gasche/random-generator/commit/74c2ea0f57d79c56328e6c6d6bf6eb2850436153), a small generator library that is parametrized over a probability monad, and can in particular compute discrete distributions. There one can check uniformity for small inputs, and I did. The fact that I completely messed up the port to QCheck, introducing several bugs, suggests that this testing strategy is not enough...)

test/core/QCheck_expect_test.ml
gasche: This is a good suggestion, I'll try to apply it shortly, thanks!

gasche: Note: for `pos_split`, testing the n-way split only would be fairly bad, because it is a very special case where the output must be [1, 1, 1, ...]. For `nat_split`, where some numbers returned can be 0, the n-way split can return arbitrary values in [0; n] in each element, so the n-way split is more representative of any other split.

test/core/QCheck_expect_test.ml
jmid: I agree about avoiding introducing CI-tests that fail with a non-zero probability.
Currently, CI-tests are run with the same seed however, which is deterministic (if not optimal).
By "stat tests" I just meant a test in `module Stats`.

While writing tuple/bind tests a few months back I added a couple of stat tests of these to confirm visually - which works nicely with expect tests: https://github.com/c-cube/qcheck/pull/176

For example, below is a (pretty IMHO) histogram and basic (deterministic 😄) average/stddev/... of the distribution of the sum of two `int`s in [0;100]. It is not a great guarantee, but still nice to confirm a reasonable average and histogram shape and the expected min and max:
https://github.com/c-cube/qcheck/blob/654abe135b21154e94e8690e0ac2b6a19e8888c8/test/core/QCheck_expect_test.expected#L507-L531

test/core/QCheck_expect_test.ml
jmid: I made a shot at this in 71675b5276372310508885986d5c41bafddce094

test/core/QCheck_expect_test.ml
jmid: I added a simple `Stats` test for `size:1` to visualize the distribution in 71675b5276372310508885986d5c41bafddce094

Issue #190, opened 2021-10-11
Deriver: support qcheck and qcheck2


vch9:
The deriver should create QCheck and QCheck2 generators.
The choice could be given in the deriver's plugin name: `[@@deriving qcheck]` and `[@@deriving qcheck2]`.

c-cube:
Would it make sense for `[@@deriving qcheck]` to also produce an arbitrary object in addition to gen + shrinker? The suffix versions would be used to derive only specific components.

Issue #189, opened 2021-10-11
Deriver: deriver in interface files


vch9:
Deriver in implementation file was done in #171, deriver in interface files is not essential but should exist at some point. 

```ocaml
(* foo.mli *)
type t 
[@@deriving qcheck]

(* should produce ==> *)
val gen : t QCheck.Gen.t
```

Niols:
I was wondering if there was a use case where the sized generator for one type would rely on the sized generator for another type even though they don't recursively depend on one another.

That being said, I'm all for small iterations and complaints-driven development, so only generating `val gen_<type>` sounds great!

PR #195, opened 2021-10-14
Deriver: share fuel between mutual generators

### LINKED ISSUE #188: Deriver: share fuel between mutual recursive types
As @jmid raised in #171. Mutual recursive generators can lead to infinite generation. 

> Regarding the mutually recursive datatypes: This limitation should probably be addressed (either now or in a separate issue). I think it could be solved like the hand-written example, by deriving a collection of fueled/sized generators for each datatype (gen_stmt_sized,gen_block_sized), letting these be the interface that drives mutually-recursive generators, and finally have a little top-level wrapper that passed them each to sized

For the following type we currently have:

```ocaml
type stmt =
  | Assign of string * string
  | Block of block
  | If of string * stmt
  | While of string * stmt

and block = Bl of stmt list
[@@deriving qcheck]

(* produces ==> *)
let rec gen_stmt () =
  QCheck.Gen.sized
  @@ QCheck.Gen.fix (fun self -> function
       | 0 ->
           QCheck.Gen.frequency
             [
               ( 1,
                 QCheck.Gen.map
                   (fun (gen0, gen1) -> Assign (gen0, gen1))
                   (QCheck.Gen.pair QCheck.Gen.string QCheck.Gen.string) );
               (1, QCheck.Gen.map (fun gen0 -> Block gen0) (gen_block ()));
             ]
       | n ->
           QCheck.Gen.frequency
             [
               ( 1,
                 QCheck.Gen.map
                   (fun (gen0, gen1) -> Assign (gen0, gen1))
                   (QCheck.Gen.pair QCheck.Gen.string QCheck.Gen.string) );
               (1, QCheck.Gen.map (fun gen0 -> Block gen0) (gen_block ()));
               ( 1,
                 QCheck.Gen.map
                   (fun (gen0, gen1) -> If (gen0, gen1))
                   (QCheck.Gen.pair QCheck.Gen.string (self (n / 2))) );
               ( 1,
                 QCheck.Gen.map
                   (fun (gen0, gen1) -> While (gen0, gen1))
                   (QCheck.Gen.pair QCheck.Gen.string (self (n / 2))) );
             ])

and gen_block () =
  QCheck.Gen.frequency
    [
      (1, QCheck.Gen.map (fun gen0 -> Bl gen0) (QCheck.Gen.list (gen_stmt ())));
    ]

let gen_stmt = gen_stmt ()
let gen_block = gen_block ()
```

Following @jmid's solution, we would have:

```ocaml
open QCheck.Gen (* for readibility purpopose *)

let rec gen_stmt n =
  match n with
  | 0 -> map2 (fun x y -> Assign (x, y)) string string
  | n ->
      oneof
        [
          map (fun x -> Block x) (gen_block (n - 1));
          map2 (fun x y -> Assign (x, y)) string string;
          map2 (fun x y -> If (x, y)) string (gen_stmt (n - 1));
          map2 (fun x y -> While (x, y)) string (gen_stmt (n - 1));
        ]

and gen_block n = map (fun x -> Bl x) (list @@ gen_stmt (n - 1))

let gen_stmt = sized @@ gen_stmt
let gen_block = sized @@ gen_block

```

vch9:
Fixes #188 

Now mutual recursive generators share the size fuel (or indicator w/e).
```ocaml
open QCheck.Gen

let rec gen_tree_sized n =
  map
    (fun gen0 -> Node gen0)
	(map
	  (fun (gen0, gen1) -> (gen0, gen1))
	  (pair int (gen_forest (n / 2))))))]

and gen_forest_sized n = match n with
  | 0 -> pure Nil
  | n ->
    frequency
	  [(1, pure Nil);
          (1,
	        (map (fun gen0 -> Cons gen0)
                  (map (fun (gen0, gen1) -> (gen0, gen1))
                  (pair (gen_tree_sized (n / 2)) (gen_forest_sized (n / 2))))))]))

let gen_tree = sized @@ gen_tree

let gen_forest = sized @@ gen_forest
```

Are you ok with that fix @jmid ?

Also in ba96a95 I removed the use of `frequency` when the list contains one element only.


jmid:
Could I ask you to rebase? (it would be nice to have a green light from CI before merging, now that it's working again) 🙏‍

vch9:
If the CI pass and you agree on the generated code, I think we can merge this @jmid.
However, I'm not fully confident with the code I wrote, I will soon write unit tests (and/or PBTs) to test generated generators.

vch9:
> If the CI pass and you agree on the generated code, I think we can merge this @jmid. However, I'm not fully confident with the code I wrote, I will soon write unit tests (and/or PBTs) to test generated generators.

I quickly found bugs writing these tests. They are not yet related to this specific PR I let you decide whether I should add those tests in this PR, or if we merge and I come up with a follow-up PR.

jmid:
> I quickly found bugs writing these tests. They are not yet related to this specific PR I let you decide whether I should add those tests in this PR, or if we merge and I come up with a follow-up PR.
> 
> (I first wrote the deriver for `QCheck.arbitrary`, I then switched to `QCheck.Gen.t`, some generators are named differently and I did not pay attention to this)

OK, makes sense. It probably makes most sense to submit those tests (and fixes) as a separate PR.

First: Sorry for leaving this hanging for some time... 😬

I took a quick look and found that there's now two copies of (what looks like) identical copies of `is_type_rec`:

https://github.com/c-cube/qcheck/blob/d0c0b1eea65cd68acde284928d47597aab2d85a4/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml#L43-L47

https://github.com/c-cube/qcheck/blob/d0c0b1eea65cd68acde284928d47597aab2d85a4/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml#L461-L465

It indicates that it would be worth it to take an extra clean-up pass over the code.

Unrelated: The comments with `n / 1` look like typos to me:
https://github.com/vch9/qcheck/blob/d0c0b1eea65cd68acde284928d47597aab2d85a4/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml#L25


I was a bit confused by the order of the code: First some basic helpers, some `sized_gen`, some `tuple`+`record`, ... then a group of `gen_from_` where `sized_gen` is used almost at the end. It could be nice to restructure it a bit to help the reader and add a few comments (e.g., about how you break all tuples down to 2-3-4 combinations).

There's already a tests for recursive types (tree, expr), for two-level (expr-value), and truly mutually recursive types (tree-forest), so you've done a great effort there already. If you feel like putting the plugin to the test, I just remembered this one
https://github.com/yomimono/ocaml-test-omp/blob/d086037027537ba4e23ce027766187979c85aa3d/test/parsetree_405.ml
which I think could make a solid stress test of the mutual recursion code! 😬 😉

Finally: I'm super happy with this deriving plugin (I started using it myself to avoid writing this boilerplate) - so thanks for your effort! 🙏‍

c-cube:
I haven't looked, but thank you @vch9 for your hard work, and @jmid for the reviewing!

question: does this use ppxlib's mechanism in such a way that `[@@deriving_inline ]` works? That'd also be super cool :)

vch9:
> I took a quick look and found that there's now two copies of (what looks like) identical copies of is_type_rec:

It does not surprise me, my rebase was suspicious.

> It indicates that it would be worth it to take an extra clean-up pass over the code.

and, yes!

> I was a bit confused by the order of the code: First some basic helpers, some sized_gen, some tuple+record, ... then a group of gen_from_ where sized_gen is used almost at the end. It could be nice to restructure it a bit to help the reader and add a few comments (e.g., about how you break all tuples down to 2-3-4 combinations).

Yes it is not ideal, I will restructure the code before this MR's changes, with more documentation obviously :). Even though, the commits in this MR are not really great: I introduced a signifcant change in d1edc6f3879cfb810e2ea918a4d36c91a511a7a0, which I generalize in a7cb9b6e92ce05bdb58cb77cdc0b163885300f14 and lead to several removals from previous commit. I think with the testing suite we have for the deriver, it could be a good thing to restart from scratch to produce a more sustainable code (I had trouble to read my own code during the rebase).

> If you feel like putting the plugin to the test, I just remembered this one
https://github.com/yomimono/ocaml-test-omp/blob/d086037027537ba4e23ce027766187979c85aa3d/test/parsetree_405.ml
which I think could make a solid stress test of the mutual recursion code! grimacing wink

I'll have a look, sounds like a solid test

vch9:
> question: does this use ppxlib's mechanism in such a way that [@@deriving_inline ] works? That'd also be super cool :)

I have not used `[@@deriving_inline]` yet in my life, but this is something I would drastically need, I'll have a look.

vch9:
@jmid This should be ready for a second round.

I introduced several commits to move/comments code which improve I believe the readability (I'm open to other suggestions as well). In particular `src/ppx_deriving_qcheck/qcheck_generators.ml` should facilitate #190.

src/ppx_deriving_qcheck/args.ml
```diff
@@ -0,0 +1,24 @@
...
+*)
+let rec curry_args ~loc args body =
+  match args with
+  | [] -> body
+  | x :: xs -> [%expr fun [%p x] -> [%e curry_args ~loc xs body]]
+
+(** [apply_args args body] applies parameters to [body]
+
+    e.g.:
+    curry_args [gen_a; gen_b] f => f gen_a gen_b
```
jmid, line 16: minor typo: `apply_args`

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,249 @@
...
+(** ppx_deriving_qcheck is a ppx deriver for QCheck generators. It does a traversal
+    map on type declarations annoted with [QCheck].
 
+    Example:
     {[
-    type tree = Leaf of int | Node of tree * tree
+    module Tree : sig
+      type t
+
+      val gen : t QCheck.Gen.t
```
jmid, line 15: typo: `val gen_t`

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,249 @@
...
+    end = struct
+      type t = Leaf | Node of int * tree * tree
```
jmid, line 17: typo: `type t = Leaf | Node of int * t * t`

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,249 @@
...
 let rec longident_to_str = function
   | Lident s -> s
   | Ldot (lg, s) -> Printf.sprintf "%s.%s" (longident_to_str lg) s
   | Lapply (lg1, lg2) ->
       Printf.sprintf "%s %s" (longident_to_str lg1) (longident_to_str lg2)
 
-let rec is_rec_typ typ_name = function
+(** [is_rec_type_decl typ_names typ] looks for elements of [typ_names]
+    recursively in [typ]. *)
+let rec is_rec_type_decl typ_names typ =
```
jmid, line 68: very minor:
these 3 mutually recursive functions could reordered
```ocaml
let rec is_rec_type ...
let is_rec_constr_decl ,,,
let is_rec_type_decl ...
```
so that only the first is recursive, no?

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -378,69 +365,131 @@ and gen_from_arrow ~loc ~env left right =
...
 
-  [%stri let [%p pat_gen] = [%e gen]]
+(** [derive_gen ~loc xs] creates generators for type declaration in [xs].
 
-let derive_gen ~loc xs : structure =
-  match xs with
-  | _, [ x ] -> [ gen_from_type_declaration ~loc x ]
+    He also has an hidden purpose: he sets the environment prior to
+    the derivation. He identifies recursive types declarations (e.g list)
+    and its external arguments (e.g. 'a list).
```
jmid, line 464: Minor (lang): *It* also ... : *it* sets ... *It* identifies  (I'm not a native English speaker either 🙂)

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -378,69 +365,131 @@ and gen_from_arrow ~loc ~env left right =
...
-  match args with
-  | [] -> body
-  | x :: xs -> [%expr fun [%p x] -> [%e curry_args ~loc xs body]]
+    It returns either `Recursive or `Normal.
 
-let gen_from_type_declaration ~loc ?(env = TypeGen.empty) td =
-  let name = td.ptype_name.txt in
-  let pat_gen = pat ~loc name in
+    - `Normal of expression:
+    The derivated generator is not recursive, we return only the generator.
```
jmid, line 387: Minor: The *derived* generator

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -378,69 +365,131 @@ and gen_from_arrow ~loc ~env left right =
...
+
+    - `Recursive of expression * expression
+    The derivated generator was recursive (i.e. val gen : n -> t Gen.t), we return
```
jmid, line 390: Minor: The *derived* generator *is*

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -378,69 +365,131 @@ and gen_from_arrow ~loc ~env left right =
...
   in
   let x, obs = aux right in
+  (* TODO: export this in qcheck_generators for https://github.com/c-cube/qcheck/issues/190 *)
   let arb = [%expr QCheck.make [%e x]] in
   [%expr
     QCheck.fun_nary QCheck.Tuple.([%e observable left] @-> [%e obs]) [%e arb]
     |> QCheck.gen]
 
-let gen_from_kind_variant ~loc ~env typ_name xs =
+let gen_from_kind_variant ~loc xs =
```
jmid, line 377: This is a pretty short function with a confusing name (there's both `gen_from_kind_invariant` and `gen_from_invariant` - and it has nothing to do with "kinds" as in type theory).
I think it would be clearer if this was just inlined in `gen_from_type_declaration` (where the other cases also take a couple of lines).

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,249 @@
...
+
+(** {2. Generator constructors} *)
+
+(** [gen lg args] creates a generator using [lg].
+
+    The longident can either be a:
+    - Lident s: We transform to gen_s (or gen if s = "t")
+    - Ldot (lg, s): We transform to qualified generator (e.g. B.gen)
+*)
+let gen ~loc lg args =
```
jmid, line 136: `gen` is not very descriptive (it's all generators).
`gen_ident` or `gen_longident` perhaps?

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,249 @@
...
+     else
+       name s |> A.evar |> Args.apply_args ~loc args
+  | Ldot (lg, s) ->
+     A.(pexp_ident (Located.mk @@ Ldot (lg, name s))) |>
+       Args.apply_args ~loc args
   | Lapply (_, _) -> raise (Invalid_argument "gen received an Lapply")
 
-let frequency ~loc l = [%expr QCheck.Gen.frequency [%e l]]
+(** [tree nodes leaves] creates a tree like generator *)
+let tree ~loc nodes leaves =
```
jmid, line 152: rename to `gen_tree`? (more descriptive - and consistent with the naming following later)

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,249 @@
...
+    type t = Leaf | Node of int * t * t
+
+    (* or *)
+
+    type t = [`Leaf | `Node of int * t * t]
+    ]}
+
+    Therefore, [is_rec] and [to_gen] are different for variants and polymorphic
+    variants. *)
+let sized ~loc (is_rec : 'a -> bool) (to_gen : 'a -> expression) (xs : 'a list) =
```
jmid, line 180: Rename to `gen_sized`?

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,249 @@
...
+    {[
+    let gen = QCheck.Gen.(map (fun (x, y) -> (x, y)) (pair int int))
+    ]}
 
+    With [f], building Foo:
+    {[
+    let gen = QCheck.Gen.(map (fun (x, y) -> Foo (x, y)) (pair int int))
+    ]}
+*)
 let tuple ~loc ?(f = fun x -> x) tys =
```
jmid, line 215: rename to `gen_tuple`?

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,249 @@
...
+    let gen = QCheck.Gen.(map (fun (x, y) -> {left = x; right = y}) (pair int int))
+    ]}
 
+    With [f], building Foo:
+    {[
+    let gen = QCheck.Gen.(map (fun (x, y) -> Foo {left = x; right = y}) (pair int int))
+    ]}
+
+*)
 let record ~loc ~gens ?(f = fun x -> x) xs =
```
jmid, line 241: rename to `gen_record`?

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,249 @@
...
+         Args.apply_args ~loc [ [%expr (n / 2)] ]
+     else
+       name s |> A.evar |> Args.apply_args ~loc args
+  | Ldot (lg, s) ->
+     A.(pexp_ident (Located.mk @@ Ldot (lg, name s))) |>
+       Args.apply_args ~loc args
   | Lapply (_, _) -> raise (Invalid_argument "gen received an Lapply")
 
-let frequency ~loc l = [%expr QCheck.Gen.frequency [%e l]]
+(** [tree nodes leaves] creates a tree like generator *)
```
jmid, line 151: I was first confused by `tree`, thinking "why are those type cases handled as a special case?"
After having read further and returning, I can now follow this conceptually (I like trees too!).
However, cut-to-the-bone: this is just a helper function.
It could be inlined in `sized` (there's morally max 2 lines of functionality)

jmid (commented):
OK. Thanks again for your efforts! I've now gone over it.

I think it really helped breaking it into a few more files 👍‍

- There are some very minor things (typos).
- some function naming still seems a bit off (I like consistency 😄)
- other functions should just be inlined IMO

Generally it is imperative with global `ref`s which saddens this old FP'er. I wonder if that can bite us at some point (parallel ppx-runs w/multicore?) 🤔

Personally I would like that the recursive generators gave rise to both `gen_t` and `gen_t_sized`:
- that leaves users the flexibility to use `Gen.sized_size` with a size-generator of their own choice
- that future proves it for when we switch generators to all be sized

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: This is not a typo, but more a design choice. If type name is `t`, I do not add a suffix to the generator name. I don't know if it's a standard convention but I've seen multiple derivers do this.

src/ppx_deriving_qcheck/args.ml
vch9: Fixed in 4a25b84dd0811e5171b131482c1f60822527ecab.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Fixed in 4a25b84dd0811e5171b131482c1f60822527ecab.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: You are right, 4eca3a41b78fdc49ea1867291afade5c75960624.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Fixed in 696b9b9aecbcdeaa75b1324f2ab75f0f813630f2.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Fixed in 696b9b9aecbcdeaa75b1324f2ab75f0f813630f2.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Fixed in 696b9b9aecbcdeaa75b1324f2ab75f0f813630f2.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: `kind_variant` and `variant` refers to nodes in the OCaml Parsetree.
image
image

But yes, the function did become relatively small, I can inline it, done in 414264448a0a0f79980e02dae760690d5dd4aed5.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: You are right: d527aa353bf3f971791ff1a3e112d8d2d24179d1

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Done in d527aa353bf3f971791ff1a3e112d8d2d24179d1

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Done in d527aa353bf3f971791ff1a3e112d8d2d24179d1

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Done in d527aa353bf3f971791ff1a3e112d8d2d24179d1.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Done in d527aa353bf3f971791ff1a3e112d8d2d24179d1

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Done in 71b284f3fb7e5105d3ce2e84733126a77c16b13c.

vch9:
> Personally I would like that the recursive generators gave rise to both gen_t and gen_t_sized:
>
>   - that leaves users the flexibility to use Gen.sized_size with a size-generator of their own choice
>   - that future proves it for when we switch generators to all be sized

It is supposed to be the case (see tests for instance). I tested with `utop` without the fix in https://github.com/c-cube/qcheck/issues/207 and I got:
```ocaml
utop # type t = Leaf of int | Node of t * t [@@deriving qcheck];;
type t = Leaf of int | Node of t * t
val gen : t QCheck.Gen.t = <fun>
```
and with the fix:
```ocaml
utop # type t = Leaf of int | Node of t * t [@@deriving qcheck];;
type t = Leaf of int | Node of t * t
val gen_sized : int -> t QCheck.Gen.t = <fun>
val gen : t QCheck.Gen.t = <fun>
```

> Generally it is imperative with global refs which saddens this old FP'er. I wonder if that can bite us at some point (parallel ppx-runs w/multicore?) thinking

The imperative can easily be removed I believe, I thought using side effects of our loved language would simplify the code. This can be changed in this PR, or whenever this cause an issue. The environment is calculated at the beginning and can be passed as a parameters in every `gen_*` functions rather than using an imperative reference.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
jmid: Sorry, my bad! 😬

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
jmid: I suggest we document the behavior in the README.md.
Here's a suggestion to insert after line 19  https://github.com/c-cube/qcheck/blob/71b284f3fb7e5105d3ce2e84733126a77c16b13c/src/ppx_deriving_qcheck/README.md?plain=1#L19
(please fill-out/check/correct the derived names and types):
```
For `type tree` we derive two generators:
- `val gen_tree : tree Gen.t` and
- `val gen_tree_sized : ...`
For non-recursive types the latter is however not derived.

For types with the name `t` (`type t = ...`) which is a common idiom in OCaml code, 
the deriver omits the name from the derived generators, thus producing `val gen : ...` and
optionally `val gen_sized : ...`. 
```

jmid:
> > Personally I would like that the recursive generators gave rise to both gen_t and gen_t_sized: [...]
> 
> It is supposed to be the case (see tests for instance). I tested with `utop` without the fix in #207 and I got:

Oh wow! Great though that we agree 😀

> > Generally it is imperative with global refs which saddens this old FP'er. I wonder if that can bite us at some point (parallel ppx-runs w/multicore?) thinking
> 
> The imperative can easily be removed I believe, I thought using side effects of our loved language would simplify the code. This can be changed in this PR, or whenever this cause an issue. The environment is calculated at the beginning and can be passed as a parameters in every `gen_*` functions rather than using an imperative reference.

I promised myself not to insist on this - merely only point it out 😅
The `curr_type ()` relies on the underlying `ref option` not being `None` though (which is how every Null-pointer bug in C/Java/... starts).
@c-cube What do you think?

Overall: Thanks a bunch! I think the code is in much clearer now.
I added a short documentation suggestion for the README.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Added this in 29d41913c2edc0a52ef35fee667698c591b2debe, thanks for the suggestion :)

vch9:
> I promised myself not to insist on this - merely only point it out sweat_smile
The curr_type () relies on the underlying ref option not being None though (which is how every Null-pointer bug in C/Java/... starts).

I think the worst case scenario would be an exception raised on `Option.get`. However, this does not bother me at all to change to an extra parameter and can be pretty quickly.

If we agree on this, I can manage to solve #189 and #190 during the day based on this MR, plus additional tests. After these three, we should be really close to a first release.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
jmid: Sorry for being unclear: can I ask you to fill in the dots `...`? 🙂

vch9: Oh well yes, I thought you meant actual dots

jmid: In principle the `...` can work.
However, I prefer being explicit about things that may seem obvious to us - because they may not be so for newcomers 🙏‍

vch9: I kept "(`type t = ...`)"

src/ppx_deriving_qcheck/README.md
```diff
@@ -20,12 +20,12 @@ let test =
 
 For `type tree` we derive two generators:
 - `val gen_tree : tree Gen.t` and
-- `val gen_tree_sized : ...`
+- `val gen_tree_sized : int -> tree Gen.t`
```
jmid, line 23: Very minor - and my mistake: Can you add a newline here?
Otherwise line 24 is rendered as part of the second list item.
Sorry, I'm still more well-versed in LaTeX than Markdown 😅

vch9: Here the fixup on fixup on commit comes !

jmid:
> > The curr_type () relies on the underlying ref option not being None though (which is how every Null-pointer bug in C/Java/... starts).
> 
> I think the worst case scenario would be an exception raised on `Option.get`. However, this does not bother me at all to change to an extra parameter and can be pretty quickly.

I think it would be nice to avoid the mutable state if we can and if it isn't a big issue to change.
I suggest the parameter name `curr_type` 😅

> If we agree on this, I can manage to solve #189 and #190 during the day based on this MR, plus additional tests. After these three, we should be really close to a first release.

I think this is already shaping up nicely! I'd be happy to release it as 0.2.1 without #189 and #190 👍‍ 
@c-cube ?

vch9:
> I think it would be nice to avoid the mutable state if we can and if it isn't a big issue to change.
I suggest the parameter name curr_type sweat_smile

I'll clean the history and change this.

> I think this is already shaping up nicely! I'd be happy to release it as 0.2.1 without #189 and #190 👍‍

I'd be in favor to wait for #190 at least.

c-cube:
note: the special case `t` is pervasive in OCaml, and in other derivers, I think @vch9 made the right  call :)

c-cube:
> > The imperative can easily be removed I believe, I thought using side effects of our loved language would simplify the code. This can be changed in this PR, or whenever this cause an issue. The environment is calculated at the beginning and can be passed as a parameters in every `gen_*` functions rather than using an imperative reference.
> 
> I promised myself not to insist on this - merely only point it out sweat_smile The `curr_type ()` relies on the underlying `ref option` not being `None` though (which is how every Null-pointer bug in C/Java/... starts). @c-cube What do you think?

what's the context? if it's global mutable state then it's kind of bad, indeed. Is it for mutually recursive functions?

vch9:
@jmid you should be happy with f6d55c65309955df494f0f4156f9a603fcba18b9! I removed references, this lead to the removal of several functions that could have been inlined :p

vch9:
> > > The imperative can easily be removed I believe, I thought using side effects of our loved language would simplify the code. This can be changed in this PR, or whenever this cause an issue. The environment is calculated at the beginning and can be passed as a parameters in every `gen_*` functions rather than using an imperative reference.
> > 
> > 
> > I promised myself not to insist on this - merely only point it out sweat_smile The `curr_type ()` relies on the underlying `ref option` not being `None` though (which is how every Null-pointer bug in C/Java/... starts). @c-cube What do you think?
> 
> what's the context? if it's global mutable state then it's kind of bad, indeed. Is it for mutually recursive functions?

It was a global mutable state to remember recursive functions and facilitate at several locations whether the type was recursive or not. However, I removed the global state for a variable moved around in f6d55c65309955df494f0f4156f9a603fcba18b9.

vch9:
Remove the `(kind ppx_rewriter)` and added `ppx_deriving` as a dependency.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,232 @@
...
+    If no recursive node is found, the type is _not_ recursive, we build a
+    generator using frequency.
 
-let tree ~loc nodes leaves =
-  [%expr
-    QCheck.Gen.sized
-    @@ QCheck.Gen.fix (fun self -> function
-         | 0 -> [%e leaves] | n -> [%e nodes])]
+    However, if recursive nodes are found, we build a tree like generator using
+    {!tree}.
```
jmid, line 145: Needs quick update as `tree` is no longer in PR code.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -381,66 +351,123 @@ and gen_from_arrow ~loc ~env left right =
...
+
+    - `Recursive of expression * expression
+    The derived generator was recursive (i.e. val gen : n -> t Gen.t), we return
+    the sized generator version, and a normal generator using this last with
+    [Gen.sized].
+*)
+let gen_from_type_declaration ~loc ~env td =
+  let (module A) = Ast_builder.make loc in
+  let ty = env.curr_type in
+  let is_rec = List.mem env.curr_type env.rec_types in
```
jmid, line 375: Since you already bind the curr_type field above, this can be written as:
```ocaml
let is_rec = List.mem ty env.rec_types in
```

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -381,66 +351,123 @@ and gen_from_arrow ~loc ~env left right =
...
-            let x = td.ptype_name.txt in
-            let gen = name x |> A.evar in
-            let expr = [%expr [%e gen] ()] in
-            TypeGen.add x expr env)
-          TypeGen.empty xs
-      in
-      let gens = List.map (gen_from_type_declaration ~loc ~env) xs in
-      mutually_recursive_gens ~loc gens
+     let typ_names = List.map (fun x -> x.ptype_name.txt) xs in
+     let env = { curr_type = ""; rec_types = []; curr_types = typ_names } in
```
jmid, line 457: `curr_type = ""` surprised me a bit at first.

In a sense it moves the "absence of `curr_type` representation from `None` to `""`.

However, we no longer have a `ref` - and here the absence is only local to `derive_gen`.
Any reader can thus convince him/her/themself that `curr_type` will be updated to be
non-empty just a few lines below (l.465) before being used.

I therefore think it is fine 👍‍ (and definitely better than the ref!)

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Great catch :) https://github.com/c-cube/qcheck/pull/195/commits/0ac00e4c992d9ff41a737eb8ffdda7a598ab9b38

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Yep, fd6821fd249e2fab9ce4b75f083e02278337a4dd

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -381,66 +351,123 @@ and gen_from_arrow ~loc ~env left right =
...
+          | [%stri let rec [%p? pat] = [%e? expr]] ->
+           A.value_binding ~pat ~expr
+        | _ -> assert false) to_mutualize_gens
+  in
+  let mutual_gens = A.pstr_value Recursive gens in
+  mutual_gens :: normal_gens
+
+(** [derive_gen ~loc xs] creates generators for type declaration in [xs].
+
+    It also has a hidden purpose: it sets the environment prior to
```
jmid, line 442: I think the hidden side-effect purpose is gone with the `env`-passing. 
If you agree, then I think the comment should be updated.

vch9: Absolutely yes 0c71322b7fac81a33afe8387bbaabe08c2938831

jmid:
> @jmid you should be happy with [f6d55c6](https://github.com/c-cube/qcheck/commit/f6d55c65309955df494f0f4156f9a603fcba18b9)! I removed references, this lead to the removal of several functions that could have been inlined :p

I know this is intended as a tongue-in-cheek remark 😅 ...but just to be clear:
I am all for helper functions ... when they help readability of the code.
I suppose I also prefer that they match entities in the domain (here: longindent, env, basic types, type constructors, ...)

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,232 @@
...
+    - Lident s: We transform to gen_s (or gen if s = "t")
+    - Ldot (lg, s): We transform to qualified generator (e.g. B.gen)
+*)
+let gen_longident ~loc ~env lg args =
   let (module A) = Ast_builder.make loc in
   match lg with
   | Lident s ->
-      Option.value ~default:(name s |> A.evar) @@ TypeGen.find_opt s env
-  | Ldot (lg, s) -> A.(pexp_ident (Located.mk @@ Ldot (lg, name s)))
+     if List.mem s env.rec_types then
```
jmid, line 127: In this case I think the code actually reads clearer with a helper function over environments:
```ocaml
if is_rec env s then
```
There are a bunch of other ad-hoc `is_rec` at the bottom which could be eliminated and instead reuse such a helper.

Ideally `type env` and these helpers could be collected in an `Env` module 😀


However there's also something weird about having parallel `is_rec*`
- for `string`s and
- for `core_type` 🤔

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,232 @@
...
+ *)
 
 let rec longident_to_str = function
   | Lident s -> s
   | Ldot (lg, s) -> Printf.sprintf "%s.%s" (longident_to_str lg) s
   | Lapply (lg1, lg2) ->
       Printf.sprintf "%s %s" (longident_to_str lg1) (longident_to_str lg2)
 
-let rec is_rec_typ typ_name = function
+let rec is_rec_typ typ_names = function
```
jmid, line 66: Now that we have `~env` I think these three functions should be updated to just pass an `env` parameter around
(where the type names can then be looked up when they need to)

Kindly check that the comments are also updated - then I promise to shut up and accept after this 😅

jmid:
Just to repeat: I think the outcome of these iterations is clearer, simpler, and shorter.
Thanks again for a great effort! 🙏‍

vch9:
> Just to repeat: I think the outcome of these iterations is clearer, simpler, and shorter.

I also believe that, do not worry :)

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: I introduced a module `Env` with the type and `is_rec`. I'm do not think I can add other helpers functions.

I believe having `is_rec` in `Env` reduces the weirdness:
* If you do `Env.is_rec` you ask if the type is recursive based on the previously computed environment
* If you do `is_rec_XXX` you ask to compute whether the type is recursive based on the known environment

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Also done, I believe I did not forget to update comments, but the PR begins to be very large

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -1,243 +1,232 @@
 open Ppxlib
+module G = Qcheck_generators
```
jmid, line 2: Perhaps use `QCheck_generators` for for consistency with `QCheck`?

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -253,43 +242,42 @@ let record ~loc ~gens ?(f = fun x -> x) xs =
...
-          tuple ~loc tys
-      | { ptyp_desc = Ptyp_constr ({ txt = ty; _ }, _); _ } ->
-          let x = TypeGen.find_opt (longident_to_str ty) env in
-          Option.value ~default:(gen ~loc ~env ty) x
-      | { ptyp_desc = Ptyp_var s; _ } -> gen ~loc (Lident s)
+          gen_tuple ~loc tys
+      | { ptyp_desc = Ptyp_constr ({ txt = ty; _ }, args); _ } ->
+         let args = List.map (gen_from_type ~loc ~env) args in
+         gen_longident ~loc ~env ty args
+      | { ptyp_desc = Ptyp_var s; _ } -> gen_longident ~loc ~env (Lident s) []
```
jmid, line 271: Insert  a newline, so that the `var` case body is consistent with `variant` and `arrow` below?

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -381,66 +347,123 @@ and gen_from_arrow ~loc ~env left right =
...
+        | _ -> assert false)
       td.ptype_params
   in
+  let (args_pat, args_expr) = List.split args in
 
   let gen =
     match td.ptype_kind with
-    | Ptype_variant xs -> gen_from_kind_variant ~loc ~env name xs
+    | Ptype_variant xs ->
+       let is_rec cd = is_rec_constr_decl env cd in
```
jmid, line 386: Just spotted  a different indentation than the following two cases

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Done, thank you

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Done, thank you for catching this

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
jmid: Looks good - and I agree it starts to be large. The diff could have been reduced by splitting it up (I don't think we should do so at this point though).
In the end the fuel-sharing fix triggered an pre-release overhaul 🤷‍♂️

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
vch9: Also done in 9d846dfe0a8d2032212c01f66bdb65b8009ec534

jmid:
Great. I think this shaped out very nicely - thanks for your effort! 😀

I'll let the CI go green before merging (which also gives @c-cube a brief window to comment/object/celebrate...)

vch9:
Rebased the history - thanks a lot for the review!

vch9:
> > If the CI pass and you agree on the generated code, I think we can merge this @jmid. However, I'm not fully confident with the code I wrote, I will soon write unit tests (and/or PBTs) to test generated generators.
> 
> I quickly found bugs writing these tests. They are not yet related to this specific PR I let you decide whether I should add those tests in this PR, or if we merge and I come up with a follow-up PR.
> 
> (I first wrote the deriver for `QCheck.arbitrary`, I then switched to `QCheck.Gen.t`, some generators are named differently and I did not pay attention to this)

Follow-up PR with those tests: https://github.com/c-cube/qcheck/pull/208

Issue #193, opened 2021-10-11
make check errors in algotest 1.5.0


olafhering:
as noted in mirage/alcotest#331, b065a81661d312fc7bbdbf95fb24229d71991a10 fails `dune runtest` with mirage/alcotest@12889aa7f993ad55434443f1c447a59f18ffc9bf

jmid:
> as noted in [mirage/alcotest#331](https://github.com/mirage/alcotest/pull/331), [b065a81](https://github.com/c-cube/qcheck/commit/b065a81661d312fc7bbdbf95fb24229d71991a10) fails `dune runtest` with [mirage/alcotest@12889aa](https://github.com/mirage/alcotest/commit/12889aa7f993ad55434443f1c447a59f18ffc9bf)

I was a bit puzzled at first since [b065a81](https://github.com/c-cube/qcheck/commit/b065a81661d312fc7bbdbf95fb24229d71991a10) is purely a documentation change.

So, this seems due to whitespace changes as far as I can see, where Alcotest 1.5.0 outputs less than 1.4.0.
Digging a bit, after a string of `grep`s in [example/alcotest/run_alcotest.sh](example/alcotest/run_alcotest.sh) there may be sequences of spaces interleaved with newlines in the result.
- stringing on `| sed 's/[ \t]*$//g' | tr -s "\n"` to strip trailing spaces and remove duplicate newlines seems to work
- one could also pass a `diff`-command to dune that would ignore whitespace differences https://dune.readthedocs.io/en/stable/concepts.html#diffing-and-promotion

jmid:
Fixed by #203

PR #202, opened 2021-12-02
[WIP] Attemps at fixing the CI

vch9:


jmid:
I believe the failure is due to alcotest changing its output #193
Accepting the new input would thus make the tests fail with a pre 1.5.0 alcotest instead.
I looked at the issue initially and proposed a fix in #193 to make the shell-script stable to such whitespace changes,
but then got side-tracked.

vch9:
We could also as it was done [in mirage packages/qcheck-alcotest/qcheck-alcotest.0.18/opam](https://github.com/mirage/alcotest/pull/331) restrict alcotest versions?

jmid:
> We could also as it was done [in mirage packages/qcheck-alcotest/qcheck-alcotest.0.18/opam](https://github.com/mirage/alcotest/pull/331) restrict alcotest versions?

Yes, that is an option.
I think it would be preferable to make our CI tests a bit more robust rather than insist on Alcotest 1.5.0 because it happens to output 1 less newline, to avoid such a small change to break it in the future. @c-cube ?

I don't find my `sed + tr` proposal particularly pretty though... 😬

c-cube:
making the test more robust is always good, yes.

jmid:
> making the test more robust is always good, yes.

@c-cube The alternative is now in PR #203. Can I ask you to merge one of these? It would be nice to get CI running again 🙂

vch9:
> > making the test more robust is always good, yes.
> 
> @c-cube The alternative is now in PR #203. Can I ask you to merge one of these? It would be nice to get CI running again slightly_smiling_face

Thank you! Thereby closing this.

PR #203, opened 2021-12-02
CI fix for Alcotest whitespace changes

jmid:
This contains the fix proposed in #193.
I have checked that it is OK locally with Alcotest 1.3.0, 1.4.0, 1.5.0. YMMV 🙃


c-cube:
fsck yeah!

PR #209, opened 2021-12-13
Deriver: qcheck and qcheck2

### LINKED ISSUE #190: Deriver: support qcheck and qcheck2

vch9:
Closes #190

This PR introduces two derivation plugins:
- `qcheck`
- `qcheck2`

The code and test is duplicated, in the vast majority it's only a renaming to `QCheck` to `QCheck2` (excepts for `fun_nary`)

```ocaml
- utop # type t = int
val gen : t QCheck.Gen.t = <fun>
val arb : t QCheck.arbitrary =
  {QCheck.gen = <fun>; print = None; small = None; shrink = None;
   collect = None; stats = []}
─( 15:46:16 )─< command 2 >─────
- utop # type t = int [@@deriving qcheck2];;
type t = int
val gen : t QCheck2.Gen.t = <abstr>
```

This it not ideal as we only derive generators then call make on them, the best would be to use arbitrary for primitive types as they also have printers, shrinkers etc.

However after #195, #208 and this one, I'm getting tired of writing boilerplate code, but I will eventually do this in the future :)


src/ppx_deriving_qcheck/QCheck_generators.ml
```diff
@@ -5,92 +5,222 @@ open Ppxlib
 
 (** {2. Type} *)
 
-let ty = Ldot (Ldot (Lident "QCheck", "Gen"), "t")
+let ty = function
+  | `QCheck -> Ldot (Ldot (Lident "QCheck", "Gen"), "t")
+  | `QCheck2 -> Ldot (Ldot (Lident "QCheck2", "Gen"), "t")
```
tmcgilchrist, line 10: Would introducing a `VERSION` module and changing these  to use a `to_module` function improve this duplication?
```
module type VERSION = sig
  type t
  val version : t
  val to_module : t -> label
end

type version = [`QCheck | `QCheck2]

let to_module = function
  | `QCheck -> "QCheck"
  | `QCheck2 -> "QCheck2"
  
```

vch9: Yes you're right, I believe we would never have more than 2 versions but that'd simplify the code a bit.

vch9: Although, I don't see the real benefit of the module type `VERSION`.
Isn't:
```ocaml
type version = [`QCheck | `QCheck2]
                    
let to_module : version -> string = function
  | `QCheck -> "QCheck"
  | `QCheck2 -> "QCheck2"

let ty version = Ldot (Ldot (Lident (to_module version), "Gen"), "t")

...

module Make (Version : sig val version : version end) = struct
```
enough?

tmcgilchrist: It would be enough, I would go with that. 
Making it a module type might be slightly tidier if more than 2 versions were expected or there were other places that needed that type. In my hacking it didn't seem worthwhile.

vch9:
Used @tmcgilchrist suggestion to reduce the code duplication and rebased to master. The commits should be squashed on merge, I'm not sure each of them compile and it's not that important if they're split.

PR #212, opened 2021-12-16
Add parameter for retrying a property while shrinking

jmid:
This PR adds a parameter `~retries` to QCheck and QCheck2 tests.
When passing a `~retries` number `n` greater than 1, the tested property is checked `n` times for each shrunk input candidate.
The default value is 1, thus causing no change in behaviour.

This feature can be useful when testing non-deterministic code with QCheck (such as https://github.com/jmid/multicoretests)
In the words of Hughes-Bolinder:Erlang11, Sec.6:
"As we explained in section 4, we ensure that tests fail when races are present simply by repeating each test a large number of times, and by running on a dual core machine. We obtained the minimal failing cases in the previous section by repeating each test 100 times during shrinking: thus we stopped shrinking a test case only when all of its candidate shrinkings passed 100 tests in a row."


c-cube:
That seems most reasonable. Would you consider adding these explanations in comments somewhere, so they live nearer the code?

c-cube (approved):
good to merge, but feel free to comment more if you want to :)

PR #214, opened 2021-12-18
Deriver fix unused var error

### LINKED ISSUE #213: Deriver: Error (warning 27 [unused-var-strict]): unused variable n.
Consider the following program:
```ocaml
open QCheck

type c =
    | A
    | B of myint [@@deriving qcheck]
and myint = int [@gen Gen.nat]
;;
Gen.generate1 gen_c
```
and a dune file:
```
(executable
 (name test)
 (libraries qcheck)
 ;;(ocamlc_flags -dsource)
 (preprocess (pps ppx_deriving_qcheck)))
```

With the latest version 1eefb2203f4cfa9c2a74be7808fa2b097f919bed this produces:
```
$ dune build test.exe
File "test.ml", lines 3-6, characters 0-30:
3 | type c =
4 |     | A
5 |     | B of myint [@@deriving qcheck]
6 | and myint = int [@gen Gen.nat]
Error (warning 27 [unused-var-strict]): unused variable n.
```

Commenting in the `-dsource` flag in the dune file and rebuilding reveals that the problem is tied to a fuel parameter:
```
$ dune build test.exe
      ocamlc .test.eobjs/byte/dune__exe__Test.{cmi,cmo,cmt} (exit 2)
[...]
open QCheck
type c =
  | A 
  | B of myint [@@deriving qcheck]
and myint = ((int)[@gen Gen.nat])
include
  struct
    let _ = fun (_ : c) -> ()
    let _ = fun (_ : myint) -> ()
    let rec gen_c_sized n =
      match n with
      | 0 -> QCheck.Gen.pure A
      | n ->
          QCheck.Gen.frequency
            [(1, (QCheck.Gen.pure A));
            (1, (QCheck.Gen.map (fun gen0 -> B gen0) gen_myint))]
    and gen_myint = Gen.nat
    let _ = gen_c_sized
    and _ = gen_myint
    let gen_c = QCheck.Gen.sized @@ gen_c_sized
    let _ = gen_c
  end[@@ocaml.doc "@inline"][@@merlin.hide ]
;;Gen.generate1 gen_c
[...]
```
@vch9 in the pattern `match` we are rebinding the (previously used) `n` parameter.
Could we solve this by simply replacing `| n ->` with `| _ ->`?
The `n` in the pattern's right-hand-side should refer to the same value - and we avoid the warning as the outer `n` is already used.

jmid:
This fixes #213 (and adds a regression test for it) with the simple patch discussed in the issue.

### DIFF
--- a/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
+++ b/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
@@ -175,7 +175,7 @@ let gen_sized ~loc (is_rec : 'a -> bool) (to_gen : 'a -> expression) (xs : 'a li
     [%expr
         match n with
         | 0 -> [%e leaves]
-        | n -> [%e nodes]
+        | _ -> [%e nodes]
     ]
 
 (** [gen_tuple ~loc ?f tys] transforms list of type [tys] into a tuple generator.
--- a/test/ppx_deriving_qcheck/deriver/test_textual.ml
+++ b/test/ppx_deriving_qcheck/deriver/test_textual.ml
@@ -385,7 +385,7 @@ let test_tree () =
        let rec gen_tree_sized gen_a n =
          match n with
          | 0 -> QCheck.Gen.pure Leaf
-         | n ->
+         | _ ->
             QCheck.Gen.frequency
               [
                 (1, QCheck.Gen.pure Leaf);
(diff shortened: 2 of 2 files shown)


vch9:
Looks good, thank you.

> Edit: Actually the deriver has a skew towards A for such "leaf cases" when the passed fuel parameter is 0. Unhappy users can however combine Gen.sized_size with a constant size-generator Gen.return 1 to avoid the skew - by always forcing gen_t_sized into the _-branch.

As you said this is not ideal but this can be a known corner-case, open to improvements by anyone

PR #215, opened 2021-12-23
Remove n fuel when its not used

### LINKED ISSUE #213: Deriver: Error (warning 27 [unused-var-strict]): unused variable n.

vch9:
#Closes #213 _again_

@jmid I went for your first option:
> I suppose it could be solved by writing a simple occurs n search (unless there is an easier way in ppxlib to check whether an identifier n occurs in a sub-expression) - and simply name it _n if it doesn't occur

I think it's pretty easy to do this kind of stuff with an `Ast_traverse`, some of the code could be refactored used this object (e.g. `is_recursive`).

### DIFF
--- a/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
+++ b/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
@@ -112,6 +112,53 @@ let is_rec_type_decl env typ =
   in
   in_type_kind || in_manifest
 
+
+(** is_n_used looks for `n` (size indication) in an expression.
+
+    For instance:
+    {[
+    type foo = A of bar | B of bar
+    and bar = Any
+    [@@deriving qcheck]
+
+    let rec gen_sized_foo n =
+      let open QCheck.Gen in
+      frequency [
+        (map (fun x -> A x) gen_bar);
+        (map (fun x -> B x) gen_bar);
+        ]
+    and gen_bar = p
+      let open QCheck.Gen in
+      pure Any
+    ]}
+
+    The type [foo] is recursive because it has a dependency to [bar] but does
+    not use the fuel as there is no "leaves" for this type.
+
+    We begin by looking for occurences of variables `n`, iff we did not find
+    any occurences, we replace `n` by `_n` in the generator's parameters. Thus,
+    avoiding an unused variable.
+ *)
+exception N_is_used
+
+class is_n_used (expr : expression) =
+object(self)
+  inherit Ast_traverse.map as super
+
+  method! expression expr =
+    match expr with
+    | [%expr n ] ->
+       raise N_is_used
+    | _ -> super#expression expr
+
+  method go () =
+    match self#expression expr |> ignore with
+    | exception N_is_used -> true
+    | () -> false
+end
+
+let is_n_used expr = (new is_n_used expr)#go ()
+
 (** {2. Generator constructors} *)
 
 (** [gen_longident lg args] creates a generator using [lg].
@@ -399,7 +446,11 @@ let gen_from_type_declaration ~loc ~env td =
     let gen = Args.curry_args ~loc args_pat gen in
     `Normal [%stri let [%p pat_gen] = [%e gen]]
   else
-    let gen = Args.curry_args ~loc (args_pat @ [A.pvar "n"]) gen in
+    let args =
+      if is_n_used gen then args_pat @ [A.pvar "n"]
+      else args_pat @ [A.pvar "_n"]
+    in
+    let gen = Args.curry_args ~loc args gen in
     let pat_gen_sized = pat ~loc ~sized:true ty in
     let gen_sized = name ~sized:true ty |> A.evar in
     let gen_normal =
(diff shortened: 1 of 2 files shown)


jmid:
Thanks for quickly taking a stab at this! 🙏‍

The proposed fix seems needlessly complex to me. IIUC, it relies on a combination of things:
- a side-effecting AST-traversal that pattern-matches on and rewrites `fun n -> ...` https://github.com/c-cube/qcheck/blob/509ef3548c5cf218b5e9094eef9267f56f7c6887/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml#L154
- an introduction of the variable `n` https://github.com/c-cube/qcheck/blob/509ef3548c5cf218b5e9094eef9267f56f7c6887/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml#L458
- a retrospective clean-up of the latter by calling the former: https://github.com/c-cube/qcheck/blob/509ef3548c5cf218b5e9094eef9267f56f7c6887/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml#L491

I think it would be cleaner to
- use a pure `is_used` interface (not performing any rewriting)
- introduce variable `n` when it `is_used` - and `_n` otherwise

since
- it tackles the underlying issue more directly
- it collects the functionality and uses less side-effects
- if we later allow overriding `gen_*_sized` (we don't presently - do we?)
  potentially there could be surprises later if an override by chance contains `fun n ->`
  (the traversal does not take variable shadowing such as `let n = ...` into account - which is fine for the derived `gen` where we are in full control - but problematic for arbitrary OCaml code)

P.s. I don't know why the Windows CI failed? 🤔

vch9:
> I think it would be cleaner to
>  * use a pure is_used interface (not performing any rewriting)
>  * introduce variable n when it is_used - and _n otherwise

Yes it would be better. On the pure-side I would still use an `Ast_mapper` the same way it is used in https://github.com/ocaml-ppx/ppxlib/blob/main/src/common.ml#L81.

tldr: use an object traversing the ast and use a local exception to detect when `n` is used in the expression

jmid:
> On the pure-side I would still use an `Ast_mapper` the same way it is used in https://github.com/ocaml-ppx/ppxlib/blob/main/src/common.ml#L81.
> 
> tldr: use an object traversing the ast and use a local exception to detect when `n` is used in the expression

Fair enough 😅 It will be the common case that `n` is used - so it makes sense to stop early on the first occurrence 👍‍

vch9:
It's done @jmid. I'm not sure the comment is placed correctly though, let me know what you think!

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -112,6 +112,53 @@ let is_rec_type_decl env typ =
   in
   in_type_kind || in_manifest
 
+
+(** is_n_used looks for `n` (size indication) in an expression.
+
+    For instance:
+    {[
+    type foo = A of bar | B of bar
+    type bar = Any
```
jmid, line 121: Typo: should read `and bar = Any`

vch9: Thank you

jmid:
> It's done @jmid. I'm not sure the comment is placed correctly though, let me know what you think!

It looks good (and simple) - thanks! 😀👍‍

PR #216, opened 2021-12-29
add ?handler parameter to check_cell_exn

gasche:
The parameter ?handler was added late to check_cell, and is missing
from its check_cell_exn counterpart for no good reason.

I hit this discrepancy when trying to debug a shrinker from within a Qcheck_alcotest test.


jmid:
LGTM 🙂

PR #217, opened 2021-12-29
Alcotest: debug_shrink

gasche:
This PR sits on top of #216; it implements a `debug_shrink` option for `QCheck_alcotest.to_alcotest`, allowing QCheck_alcotest users to debug their shrinkers.

I would consider this an important feature for QCheck_alcotest users: it's not possible in practice to develop your own shrinkers if you cannot debug the shrinker.

### DIFF
--- a/src/alcotest/QCheck_alcotest.mli
+++ b/src/alcotest/QCheck_alcotest.mli
@@ -12,7 +12,10 @@
 *)
 
 val to_alcotest :
-  ?verbose:bool -> ?long:bool -> ?rand:Random.State.t ->
+  ?colors:bool -> ?verbose:bool -> ?long:bool ->
+  ?debug_shrink:(out_channel option) ->
+  ?debug_shrink_list:(string list) ->
+  ?rand:Random.State.t ->
   QCheck2.Test.t -> unit Alcotest.test_case
 (** Convert a qcheck test into an alcotest test
     @param verbose used to print information on stdout (default: [verbose()])
--- a/example/alcotest/QCheck_alcotest_test.ml
+++ b/example/alcotest/QCheck_alcotest_test.ml
@@ -55,6 +55,6 @@ let () =
     List.map QCheck_alcotest.to_alcotest
       [ passing; failing; error; simple_qcheck; passing_tree_rev ]
   in
-  A.run "my test" [
-    "suite", suite
-  ]
+  A.run ~show_errors:true "my test" [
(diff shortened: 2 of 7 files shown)


gasche:
I rebased this PR on master now that #216 has been merged (thanks!).

vch9:
I havent had a look at the implementation but I saw your tests and examples and I strongly agree that we need this.

c-cube:
This looks good, I'm just wondering about the `shrink_debug_list`. I think it could be a predicate `string -> bool` on names instead, so it's a bit more orthogonal to the runner (easier to say "always yes" for example).

gasche:
I have mirrored the API used in `Qcheck_base_runner.run_tests`, and I think that the two APIs should always mirror each other. I certainly agree that the `debug_shrink_list` API is not optimal (currently `[]` is special-cased to mean "always yes"), but this could be improved separately.

c-cube:
Ok then :).
Maybe worth opening an issue for this to be improved later.

src/alcotest/QCheck_alcotest.mli
```diff
@@ -12,7 +12,10 @@
 *)
 
 val to_alcotest :
-  ?verbose:bool -> ?long:bool -> ?rand:Random.State.t ->
+  ?colors:bool -> ?verbose:bool -> ?long:bool ->
+  ?debug_shrink:(out_channel option) ->
+  ?debug_shrink_list:(string list) ->
+  ?rand:Random.State.t ->
   QCheck2.Test.t -> unit Alcotest.test_case
```
jmid, line 19: Even though already merged:
Could I ask you to document the new params in the .mli @gasche - pretty please? 😀🙏‍

gasche: Sure! But then again, those are the same parameters as [QCheck_base_runner.run_tests](http://c-cube.github.io/qcheck/0.18/qcheck-core/QCheck_base_runner/index.html#val-run_tests). (I can just add a cross-reference in the documentation.)

jmid: A cross-reference is fine! 
I wrote that documentation for `run_tests` only a few months back since it was missing, hence my knee-jerk missing-doc reaction 😅 I missed that empty-list was special cased though... 😬

gasche: Done in #221

PR #220, opened 2022-01-07
QCHECK_LONG_FACTOR environment variable

vch9:
In https://github.com/c-cube/qcheck/pull/142 we introduced a `QCHECK_COUNT` environment variable. It is very useful to debug flaky tests by running a dedicated pipeline with a lot higher `QCHECK_COUNT`. We decided that the `~count` in a `Test.make` overrides the environment variable (which I think is fine), but, there is no easy way to conditionally increase these given counts.

I propose another environment variable `QCHECK_COUNT_MULTIPLIER` to increase the `~count` value.

### DIFF
--- a/src/core/QCheck2.ml
+++ b/src/core/QCheck2.ml
@@ -1459,10 +1459,11 @@ module Test = struct
     }
 
   let make_cell_from_QCheck1 ?(if_assumptions_fail=default_if_assumptions_fail)
-      ?(count) ?(long_factor=1) ?max_gen
+      ?(count) ?long_factor ?max_gen
       ?(max_fail=1) ?(retries=1) ?(name=fresh_name()) ~gen ?shrink ?print ?collect ~stats law
     =
     let count = global_count count in
+    let long_factor = global_long_factor long_factor in
     (* Make a "fake" QCheck2 arbitrary with no shrinking *)
     let fake_gen = Gen.make_primitive ~gen ~shrink:(fun _ -> Seq.empty) in
     let max_gen = match max_gen with None -> count + 200 | Some x->x in
--- a/src/core/QCheck2.ml
+++ b/src/core/QCheck2.ml
@@ -1421,7 +1421,7 @@ module Test = struct
     | (_, Some x) -> int_of_string x
     | (None, None) -> default_count
   in
-  if count < 0 then invalid_arg ("count must be > 0 but value is " ^ string_of_int count) else count
+  if count <= 0 then invalid_arg ("count must be > 0 but value is " ^ string_of_int count) else count
 
   let fresh_name =
     let r = ref 0 in
--- a/src/core/QCheck2.ml
+++ b/src/core/QCheck2.ml
@@ -1415,13 +1415,19 @@ module Test = struct
 
   let default_count = 100
 
-  let global_count count =
-    let count = match (count, Sys.getenv_opt "QCHECK_COUNT") with
+  let default_long_factor = 1
+
+  let global_positive_var default env_var var =
+    let var = match (var, Sys.getenv_opt env_var) with
     | (Some x, _) -> x
     | (_, Some x) -> int_of_string x
-    | (None, None) -> default_count
+    | (None, None) -> default
   in
-  if count <= 0 then invalid_arg ("count must be > 0 but value is " ^ string_of_int count) else count
+  if var <= 0 then invalid_arg (env_var ^ " must be > 0 but value is " ^ string_of_int var) else var
+
+  let global_count count = global_positive_var default_count "QCHECK_COUNT" count
+
+  let global_long_factor long_factor = global_positive_var default_long_factor "QCHECK_LONG_FACTOR" long_factor
 
   let fresh_name =
     let r = ref 0 in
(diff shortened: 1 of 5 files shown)


c-cube:
just a remark, there is already the notion of `long` tests, which is a multiplier. Is this different? (the env var is a good idea anyway, mind you).

vch9:
> just a remark, there is already the notion of long tests, which is a multiplier. Is this different? (the env var is a good idea anyway, mind you).

I was not aware of `long` tests, what I proposed is not different. I'll change to an env variable `QCHECK_LONG_FACTOR` yes

src/core/QCheck2.ml
```diff
@@ -1394,13 +1394,19 @@ module Test = struct
...
     | (_, Some x) -> int_of_string x
-    | (None, None) -> default_count
+    | (None, None) -> default
   in
-  if count <= 0 then invalid_arg ("count must be > 0 but value is " ^ string_of_int count) else count
+  if var <= 0 then invalid_arg (env_var ^ " must be > 0 but value is " ^ string_of_int var) else var
+
+  let global_count count = global_positive_var default_count "QCHECK_COUNT" count
+
+  let global_long_factor long_factor = global_positive_var 1 "QCHECK_LONG_FACTOR" long_factor
```
vch9, line 1409: use `default_long_factor` here

vch9:
Hi @c-cube and/or @jmid :).
I've rebased this branch and add additional tests. We'd be interesting on using this for the Tezos CI, do you think you could have a look?

src/core/QCheck2.ml
```diff
@@ -1430,10 +1436,11 @@ module Test = struct
   let default_if_assumptions_fail = `Warning, 0.05
 
   let make_cell ?(if_assumptions_fail=default_if_assumptions_fail)
-      ?(count) ?(long_factor=1) ?max_gen
+      ?(count) ?long_factor ?max_gen
       ?(max_fail=1) ?(retries=1) ?(name=fresh_name()) ?print ?collect ?(stats=[]) gen law
     =
     let count = global_count count in
+    let long_factor = global_long_factor long_factor in
```
jmid, line 1443: I would have expected similar changes below in `make_cell_from_QCheck1`.
Otherwise I suspect the QCHECK_LONG_FACTOR environment variable will only work for QCheck2 tests - but not for QCheck tests.

jmid (changes requested):
Overall this is a very reasonable request of being able to configure `long_factor` through an environment variable. Code looks good and clean too. My heartfelt thanks (really!) for including tests of the new functionality! :pray: 

If I'm not mistaken, this only adds the functionality for QCheck2 - which leaves us in a bit of a limbo. Could I persuade you to copy-paste the functionality so that it will also work for QCheck tests? (the unit-tests for QCheck should largely also just be a copy-paste job)

Finally, you are also welcome to add a CHANGELOG entry for the new functionality.

vch9:
> Overall this is a very reasonable request of being able to configure `long_factor` through an environment variable. Code looks good and clean too. My heartfelt thanks (really!) for including tests of the new functionality! pray
> 
> If I'm not mistaken, this only adds the functionality for QCheck2 - which leaves us in a bit of a limbo. Could I persuade you to copy-paste the functionality so that it will also work for QCheck tests? (the unit-tests for QCheck should largely also just be a copy-paste job)
> 
> Finally, you are also welcome to add a CHANGELOG entry for the new functionality.

I did not manually test or even look deep but since a QCheck cell maker internally uses QCheck2 cell maker wouldn't this PR makes this available for QCheck as well?

src/core/QCheck2.ml
jmid: `QCheck.Test.make_cell` calls `QCheck2.Test.make_cell_from_QCheck1` to map all QCheck(1) tests to QCheck2 tests, hence my concern.

vch9: Thanks you're right, that was a mistake I thought it was covered (but tbh I did not look very deeply)

vch9:
> Overall this is a very reasonable request of being able to configure `long_factor` through an environment variable. Code looks good and clean too. My heartfelt thanks (really!) for including tests of the new functionality! pray
> 
> If I'm not mistaken, this only adds the functionality for QCheck2 - which leaves us in a bit of a limbo. Could I persuade you to copy-paste the functionality so that it will also work for QCheck tests? (the unit-tests for QCheck should largely also just be a copy-paste job)
> 
> Finally, you are also welcome to add a CHANGELOG entry for the new functionality.

You're absolutely right.
* I made the changes needed for QCheck in 8f3d032d06578dfc1c711ccbdbcb62612ff73a8a
* Updated the changelog in 5b30869ac8c5eac098bfcf92a5d3235964691f09, I also agree that it is a good thing to update the changelog every time we add a feature

src/core/QCheck2.ml
vch9: Should be fine like this

jmid (approved):

jmid:
Great, thank!
Extra credit for also adding a corresponding `TestCount` module for the QCheck(1) unit tests, yay! :clap: :smiley:

PR #223, opened 2022-01-17
QCheck2.Gen: enforce naming consistency

### LINKED ISSUE #162: `QCheck2.Gen` design considerations
With `QCheck2` still not yet released (and still being tested) we have a brief opportunity to discuss the design of its `Gen` combinators.

**Optional parameters**
---------------------------------
We have had problems with optional parameters in the past (see #75). These may cause problems when they are the only parameters. For QCheck this could usually be solved with type annotations:
```ocaml
utop # QCheck.Gen.small_string;;
- : ?gen:char QCheck.Gen.t -> string QCheck.Gen.t = <fun>
utop # QCheck.Gen.([small_string; string]);;
- : (?gen:char QCheck.Gen.t -> string QCheck.Gen.t) list = [<fun>; <fun>]
utop # (QCheck.Gen.([small_string; string]) : string QCheck.Gen.t list);;
- : string QCheck.Gen.t list = [<fun>; <fun>]
utop # QCheck.make QCheck.Gen.small_string;;
- : string QCheck.arbitrary =
{QCheck.gen = <fun>; print = None; small = None; shrink = None; collect = None;
 stats = []}
```
For `Check2` however because `QCheck2.Gen.t` is opaque, even with the fix in PR #161, the type annotation trick doesn't work anymore:
```ocaml
utop # QCheck2.Gen.small_string;;
- : ?gen:char Gen.t -> string Gen.t = <fun>
utop # QCheck2.Gen.([small_string; string]);;
Error: This expression has type string t but an expression was expected of type
         ?gen:char t -> string t
utop #  (QCheck2.Gen.([small_string; string]) : string QCheck2.Gen.t list);;
Error: This expression has type ?gen:char t -> string t
       but an expression was expected of type string t
utop # QCheck2.Test.make QCheck2.Gen.small_string;;
Error: This expression has type ?gen:char Gen.t -> string Gen.t
       but an expression was expected of type 'a Gen.t
```
Effectively, this means the (labeled) parameter is required, not optional.
As such, we should seriously consider changing `QCheck2.Gen.small_string` to one of the following:
```ocaml
  val small_string : string t                   (* just use char *)
  val small_string : char t -> string t         (* make it a regular argument *) 
  val small_string : gen:char t -> string t     (* make it an actual labelled argument *)
  ...
```
Of the three above, I would prefer one of the first two. 
Based on our choice, we should consider whether to update `string_size` accordingly following the principle of least surprise:
```ocaml
```

Looking at the `Gen` module's signature I can see a similar problem with `Gen.pint : ?origin:int -> int t` which we also need to address:
```ocaml
utop # QCheck2.Test.make QCheck2.Gen.pint;;
Error: This expression has type ?origin:int -> int Gen.t
       but an expression was expected of type 'a Gen.t
```
Here, we probably should go with a regular or labelled argument. Again, from the principle of least surprise,
we should consider updating the remaining `origin`-taking combinators to take the same kind of `origin` argument:
```ocaml
  val int_range : ?origin:int -> int -> int -> int t
  val float_bound_inclusive : ?origin:float -> float -> float t
  val float_bound_exclusive : ?origin:float -> float -> float t
  val float_range : ?origin:float -> float -> float -> float t
  val char_range : ?origin:char -> char -> char -> char t
```

Besides the`Gen.generate*` bindings, only two bindings now stand a bit out:
```ocaml
  val opt : ?ratio:float -> 'a t -> 'a option t
  val make_primitive : gen:(Random.State.t -> 'a) -> shrink:('a -> 'a Seq.t) -> 'a t
```
I think for `opt` the use of an optional parameter is warranted.
For `make_primitive` we may consider whether labelled arguments are needed.

Generator names:
-------------------------
Naming-wise, we should consider the consistency. Here are the string generators:
```ocaml
  val string : string t
  val string_of : char t -> string t
  val string_readable : string t
```
Wouldn't `string_small` be nice and consistent - and easier to find using tab-completion?

I think the same goes for the list generators (`small_list` -> `list_small`):
```ocaml
  val list : 'a t -> 'a list t
  val small_list : 'a t -> 'a list t
  val list_size : int t -> 'a t -> 'a list t
  val list_repeat : int -> 'a t -> 'a list t
```
and the array generators  (`small_array` -> `array_small`):
```ocaml
  val array : 'a t -> 'a array t
  val array_size : int t -> 'a t -> 'a array t
  val small_array : 'a t -> 'a array t
  val array_repeat : int -> 'a t -> 'a array t
```

For `int*`, `float`, and `char` generators it is harder...

Use of underscore in name suffixes
-----------------------------------------------
I spotted an unfortunate variation in the use of `_l`, `_a` or `l`, `a` suffixes:
```ocaml
  val oneof : 'a t list -> 'a t
  val oneofl : 'a list -> 'a t
  val oneofa : 'a array -> 'a t
  val frequency : (int * 'a t) list -> 'a t
  val frequencyl : (int * 'a) list -> 'a t
  val frequencya : (int * 'a) array -> 'a t
  val shuffle_a : 'a array -> 'a array t
  val shuffle_l : 'a list -> 'a list t
  val shuffle_w_l : (int * 'a) list -> 'a list t
  ...
  val flatten_l : 'a t list -> 'a list t
  val flatten_a : 'a t array -> 'a array t
  val flatten_opt : 'a t option -> 'a option t
  val flatten_res : ('a t, 'e) result -> ('a, 'e) result t
```

I know a lot of this is legacy, but with a clean `QCheck2` now is a good time to consider cleaning these things a bit up.

vch9:
Partially close #162 (the optional parameters are not adressed)

As QCheck2 is already released, I went with a "deprecated" phase, this can be back ported to QCheck1 as well.

The generator naming convention is basically `<type>_suffixes` where `_suffixes` is optional.


test/core/QCheck2_expect_test.ml
```diff
@@ -481,8 +481,8 @@ module Function = struct
       Gen.(quad  (* string -> int -> string *)
              (fun2 ~print:Print.string Observable.string Observable.int (small_string ~gen:char))
              (small_string ~gen:char)
```
jmid, line 483: also switch to `string_small`? (two occurrences)

test/core/QCheck2_expect_test.ml
```diff
@@ -420,7 +420,7 @@ module Function = struct
     Test.make ~name:"fail_pred_map_commute" ~count:100 ~long_factor:100
       ~print:Print.(triple (list int) Fn.print Fn.print)
       Gen.(triple
-             (small_list small_int)
+             (small_list nat_small)
```
jmid, line 423: Use `list_small`?

test/core/QCheck2_expect_test.ml
```diff
@@ -40,11 +40,11 @@ module Overall = struct
   let passing =
     Test.make ~name:"list_rev_is_involutive" ~count:100 ~long_factor:100
       ~print:Print.(list int)
-      Gen.(list small_int) (fun l -> List.rev (List.rev l) = l)
+      Gen.(list nat_small) (fun l -> List.rev (List.rev l) = l)
 
   let failing =
     Test.make ~name:"should_fail_sort_id" ~count:10 ~print:Print.(list int)
-      Gen.(small_list small_int) (fun l -> l = List.sort compare l)
+      Gen.(small_list nat_small) (fun l -> l = List.sort compare l)
```
jmid, line 47: Use `list_small`?

src/core/QCheck2.ml
```diff
@@ -467,21 +477,21 @@ module Gen = struct
...
 
-  (* NOTE: we keep this alias to not break code that uses [small_int]
-     for sizes of strings, arrays, etc. *)
-  let small_int = small_nat
-
-  let small_signed_int : int t = fun st ->
+  let int_small_signed : int t = fun st ->
     if RS.bool st
     then small_nat st
     else (small_nat >|= Int.neg) st
```
jmid, line 491: Use `nat_small`? (two occurrences)

src/core/QCheck2.ml
```diff
@@ -685,11 +697,14 @@ module Gen = struct
 
   let string_printable = string_size ~gen:printable nat
 
-  let small_string ?gen st = string_size ?gen small_nat st
+  let string_small ?gen st = string_size ?gen small_nat st
```
jmid, line 700: Use `nat_small`?

src/core/QCheck2.ml
```diff
@@ -685,11 +697,14 @@ module Gen = struct
...
+  let small_string = string_small
 
-  let small_list gen = list_size small_nat gen
+  let list_small gen = list_size small_nat gen
```
jmid, line 703: same here

src/core/QCheck2.ml
```diff
@@ -685,11 +697,14 @@ module Gen = struct
...
+  let small_list = list_small
 
-  let small_array gen = array_size small_nat gen
+  let array_small gen = array_size small_nat gen
```
jmid, line 706: and here

src/core/QCheck2.mli
```diff
@@ -319,11 +394,22 @@ module Gen : sig
...
+  (** Builds a string generator, length is {!small_nat}.
+      Accepts an optional character generator (the default is {!char}).
+
+      Shrinks on the number of characters first, then on the characters.
+
+      @since NEXT_RELEASE
+  *)
+
   val small_string : ?gen:char t -> string t
   (** Builds a string generator, length is {!small_nat}.
```
jmid, line 407: `nat_small`?

src/core/QCheck2.mli
```diff
@@ -319,11 +394,22 @@ module Gen : sig
 
       @since 0.11 *)
 
+  val string_small : ?gen:char t -> string t
+  (** Builds a string generator, length is {!small_nat}.
```
jmid, line 398: `nat_small`

jmid:
Thanks for taking a stab at this! Overall this LGTM 😀

I spotted a few occurrences of `small_nat` in the interface and in the implementation.
It would make sense to switch to the new names internally in the module, IMO.

There's also one remaining in line 62 of the interface in the documentation example at the top:
```
Gen.(list small_nat)
```
I'm guessing there are probably more that I missed.
Could I ask you to take an extra pass over both files to make sure we've caught (most of) them? 🙏‍

Looking at the new interface from a distance we still have a bit of inconsistency regarding signed generators:
```ocaml
  val int : int t
  val pint : ?origin : int -> int t
  val int_neg : int t
  val int_small_signed : int t
  val int_small_corners : unit -> int t
  val int_pos_corners : int list
  val int_corners : int list
```
compared to:
```ocaml
  val float : float t
  val float_p : float t
  val float_n : float t
```
One option would be to rename
- `pint` to `int_pos`
- `float_p` to `float_pos`
- `float_n` to `float_neg`

but if the majority prefers `_p` and `_n`suffixes I'm also fine with that - as long as it is consistent! 😅😉

The `int_small_signed` is a legacy from `QCheck` IIRC. Since we are renaming for consistency `nat` to me indicates non-negative and `int` indicates both positive and negative numbers. As such, while we are at it I would favor a rename
- `int_small_signed` to `int_small`

@c-cube ?

Finally, there's the "optional parameters that aren't so optional" that you didn't intend to address in this PR:
We should do so as part of rewamping the QCheck2 API IMO - but perhaps as a separate PR (before a release, please),
as there are still generators with a single not-so-optional parameter remaining in the revised interface.

vch9:
> Thanks for taking a stab at this! Overall this LGTM grinning
> 
> I spotted a few occurrences of `small_nat` in the interface and in the implementation. It would make sense to switch to the new names internally in the module, IMO.
> 
> There's also one remaining in line 62 of the interface in the documentation example at the top:
> 
> ```
> Gen.(list small_nat)
> ```
> 
> I'm guessing there are probably more that I missed. Could I ask you to take an extra pass over both files to make sure we've caught (most of) them? 🙏‍

There is most certainly other occurences, thanks for noticing (some of) them. I will correct them and have a second look.

> Looking at the new interface from a distance we still have a bit of inconsistency regarding signed generators:
> 
> ```ocaml
>   val int : int t
>   val pint : ?origin : int -> int t
>   val int_neg : int t
>   val int_small_signed : int t
>   val int_small_corners : unit -> int t
>   val int_pos_corners : int list
>   val int_corners : int list
> ```
> 
> compared to:
> 
> ```ocaml
>   val float : float t
>   val float_p : float t
>   val float_n : float t
> ```
> 
> One option would be to rename
> 
>     * `pint` to `int_pos`
> 
>     * `float_p` to `float_pos`
> 
>     * `float_n` to `float_neg`
> 
> 
> but if the majority prefers `_p` and `_n`suffixes I'm also fine with that - as long as it is consistent! sweat_smilewink

I also do prefer `_pos` to be honest. 

> The `int_small_signed` is a legacy from `QCheck` IIRC. Since we are renaming for consistency `nat` to me indicates non-negative and `int` indicates both positive and negative numbers. As such, while we are at it I would favor a rename
> 
>     * `int_small_signed` to `int_small`
> 
> 
> @c-cube ?
> 
> Finally, there's the "optional parameters that aren't so optional" that you didn't intend to address in this PR: We should do so as part of rewamping the QCheck2 API IMO - but perhaps as a separate PR (before a release, please), as there are still generators with a single not-so-optional parameter remaining in the revised interface.

I think the way to go for optional parameters would be like @sir4ur0n mentioned:
> n my opinion, whenever it makes sense, we should provide a generator that does not take any additional input (make it dead-easy to get started on QCheck) so I'd go with val small_string : string t by default.
> Then we should also provide another function that takes a mandatory - labeled or not, this depends if there are other arguments - generator.

But for instance with `opt`, I did not know what was more convenient as a suffix:
```ocaml
val opt : 'a t -> 'a option t
val opt_ratio : 'a t -> 'a option t
```

For other types such as `list`, `string`, `_of` can be an appropriate suffix I think.

vch9:
I think `_pos` is more convenient: we have a lot of suffixes in the module. For some it's obvious (e.g. `_neg`) but its more complicated for `_big` for instance.

vch9:
Regarding `make_primitive` I think we could change to:
```
val make_primitive : ?shrink : ('a -> 'a Seq.t) -> (Random.State.t -> 'a) -> 'a t
```
where `?(shrink = fun _ -> Seq.empty)`

jmid:
> > Finally, there's the "optional parameters that aren't so optional" that you didn't intend to address in this PR: We should do so as part of rewamping the QCheck2 API IMO - but perhaps as a separate PR (before a release, please), as there are still generators with a single not-so-optional parameter remaining in the revised interface.
> 
> I think the way to go for optional parameters would be like @sir4ur0n mentioned:
> 
> > n my opinion, whenever it makes sense, we should provide a generator that does not take any additional input (make it dead-easy to get started on QCheck) so I'd go with val small_string : string t by default.
> > Then we should also provide another function that takes a mandatory - labeled or not, this depends if there are other arguments - generator.
> 
> But for instance with `opt`, I did not know what was more convenient as a suffix:
> 
> ```ocaml
> val opt : 'a t -> 'a option t
> val opt_ratio : ratio:float -> 'a t -> 'a option t
> ```
I like `opt_ratio` - it is a very saying name!  😀
I also agree with @sir4ur0n that providing simple, non-parameterized generators is a good idea to lower the barrier to entry.

I guess a second implicit design principle is to try to align generator names with OCaml's type names and type constructors:
`unit`, `bool`, `char`, `sttring`, ... `int list`, `char array`
For the latter we should stick with the prefixed generator names `list int`, `array char`, ... 
If we follow this principle, we should rename `opt` to `option` though... 🤔

> For other types such as `list`, `string`, `_of` can be an appropriate suffix I think.

To keep alignment with the types and type constructors `list` and `array` should be kept as is,
as the user has to provide an element type (generator).
For `string` both the above design principles suggest to keep an un-parameterized version.

I agree about the `_of` suffix for, e.g., for `string`!
As I mentioned in https://github.com/c-cube/qcheck/issues/162#issuecomment-901294838
this makes for very readable generators in test specifications like `Gen.(string_of printable)`, `Gen.(string_of numeral), ...
We should add additional char generators, like `letter`, `ascii`, `latin1`, ... (but that is for another issue/PR! 😀)

vch9:
While addressing the suffixes and optional parameters (except `make_primitive` yet) I also looked for occurrences with `git grep <old_gen>`. I might have removed the vast majority but its a error-prone process. Once we have released `NEXT_RELEASE` we will be able to totally wipe the deprecated generators and the compilation will help us.

jmid:
> Regarding `make_primitive` I think we could change to:
> ```
> val make_primitive : ?shrink : ('a -> 'a Seq.t) -> (Random.State.t -> 'a) -> 'a t
> ```
> where `?(shrink = fun _ -> Seq.empty)`

So making `shrink` optional and removing the `gen`-label? 🤔

The QCheck parallel is probably `QCheck.make` that has a similar interface: a bunch of optional parameters (shrink, print, ...) and one unlabeled required one (the `Gen.t` generator).

- On the one hand it will be more consistent with the interface of the remaining generator combinators.
- On the other hand code of a custom QCheck2 generator could be a bit harder to read, e.g., if it passes two large arguments ("what is this second argument?").

An alternative would be to have an optional `shrink` and a required+labeled `gen`.
I don't feel strongly for either though. @c-cube ?

src/core/QCheck2.mli
```diff
@@ -161,10 +161,28 @@ module Gen : sig
       Shrinks towards [0].
   *)
 
+  val nat_origin : int -> int t
+  (** Generates non-strictly positive integers uniformly ([0] included).
+
+      Shrinks towards [origin] if specified, otherwise towards [0].
```
jmid, line 167: `[nat_origin b] shrinks towards [b]`.

src/core/QCheck2.mli
```diff
@@ -161,10 +161,28 @@ module Gen : sig
...
+
+      @since NEXT_RELEASE *)
+
   val pint : ?origin : int -> int t
   (** Generates non-strictly positive integers uniformly ([0] included).
 
-      Shrinks towards [origin] if specified, otherwise towards [0]. *)
+      Shrinks towards [origin] if specified, otherwise towards [0].
+
+      @deprecated use {!int_origin} *)
```
jmid, line 176: `use {!nat_origin}`

c-cube:
not very fond of required+labeled in this case, but why not . Remember you also need a `unit` at the end to keep the optional parameter optional.

For `int_small_signed`, yes, remove `_signed`. It should be in the docstring but the `nat` vs `int` split is clear enough I think.

src/core/QCheck2.ml
```diff
@@ -353,15 +361,20 @@ module Gen = struct
...
+
+  let option_ratio ~ratio (gen : 'a t) : 'a option t = fun st ->
     let p = RS.float st 1. in
     if p < (1. -. ratio)
     then Tree.pure None
     else Tree.opt (gen st)
 
+  let option gen = option_ratio ~ratio:0.85 gen
+
+  let opt ?(ratio = 0.85) = option_ratio ~ratio
```
jmid, line 376: Does this compile without the dreaded "this-optional-argument-is-not-so-option" warning/error?
(I realize it is curried, so that `option_ratio` will expect a later `gen` argument)

src/core/QCheck2.ml
```diff
@@ -353,15 +361,20 @@ module Gen = struct
...
   let (--.) low high = float_range ?origin:None low high
 
-  let neg_int : int t = nat >|= Int.neg
+  let int_neg : int t = nat >|= Int.neg
 
-  (** [opt gen] shrinks towards [None] then towards shrinks of [gen]. *)
-  let opt ?(ratio : float = 0.85) (gen : 'a t) : 'a option t = fun st ->
+  let neg_int = int_neg
+
+  let option_ratio ~ratio (gen : 'a t) : 'a option t = fun st ->
```
jmid, line 368: The interface in .mli doesn't use a label for `gen` AFAICS. I'm curious why this wasn't caught be the type-checker... 🤔
That however suggests that we should add a test for these new ones to ensure we don't roll out any broken signatures in the new interface... (it was when writing such tests I encountered the optional-argument issue)

src/core/QCheck2.ml
```diff
@@ -399,14 +412,16 @@ module Gen = struct
       let right = RS.bits st in
       left lor middle lor right
 
-  let pint ?(origin : int = 0) : int t = fun st ->
+  let nat_origin origin : int t = fun st ->
     let x = pint_raw st in
```
jmid, line 416: I suggest renaming `pint_raw` to `nat_pos_raw` while we are at it.
I know it isn't visible from the outside, however using the naming principles internally should make reading the implementation easier for ourselves and others going forward.

src/core/QCheck2.ml
```diff
@@ -420,15 +435,15 @@ module Gen = struct
...
+    then nat_origin 0 >|= (fun n -> - n - 1)
+    else nat_origin 0
 
   let int_bound (n : int) : int t =
     if n < 0 then invalid_arg "Gen.int_bound";
     fun st ->
       if n <= (1 lsl 30) - 2
       then Tree.make_primitive (fun a () -> Shrink.int_towards 0 a ()) (RS.int st (n + 1))
-      else Tree.map (fun r -> r mod (n + 1)) (pint st)
+      else Tree.map (fun r -> r mod (n + 1)) (nat st)
```
jmid, line 446: I think suspect this renaming introduces a bug: the else branch handles `n` bigger than 30-bit,
which will then need the old `pint` (now: `nat_origin`) to stitch together 3 calls.
Using plain `nat` outputs max 10.000 though. A statistics test should reveal the change in distribution!
This would be a good opportunity to add it 😀

src/core/QCheck2.ml
```diff
@@ -420,15 +435,15 @@ module Gen = struct
   let int : int t =
     bool >>= fun b ->
     if b
-    then pint ~origin:0 >|= (fun n -> - n - 1)
-    else pint ~origin:0
+    then nat_origin 0 >|= (fun n -> - n - 1)
+    else nat_origin 0
 
   let int_bound (n : int) : int t =
```
jmid, line 441: For consistency, should this really be called `nat_bound`? 😬

jmid:
Hm. I'm starting to realize that this has opened a can of worms and wonder if the design is going in the direction we intend it to...

In a sense, having `nat`-generators is short, user-friendly to write, and clear but it also goes against consistently using a `type`-prefix. I'm not sure replacing `nat` with `int_pos` in these will be better though:
```ocaml
  val nat_origin : int -> int t
  val nat_small : int t
  val nat : int t
  val nat_big : int t
```
I noticed that
- `nat` is non-uniform and at most 10.000 - while `int` is uniform over all OCaml `int`s.
  (one should use `nat_origin 0` to get a behavior mirroring `int`) 
- there's no `int_big`

when comparing to these:
```ocaml
  val int : int t
  val int_neg : int t
  val int_small : int t
  val int_small_corners : unit -> int t
  val int_range : ?origin:int -> int -> int -> int t
  val int_pos_corners : int list
  val int_corners : int list
```
So we end up having `int_neg` - but no `int_pos` - so we should rename `int_pos_corners` to `nat_corners` I guess?
Furthermore, it seems we are missing a generator of `int_pos/nat_(small_)corners` that utilizes `int_pos_corners`... 😬🤷‍♂️

src/core/QCheck2.ml
vch9: This is not a label for gen, it's just annotated with the type. But yes, the colors are tricky, I had to check in the source file ahah

vch9:
> I noticed that
>  * nat is non-uniform and at most 10.000 - while int is uniform over all OCaml ints.
    (one should use nat_origin 0 to get a behavior mirroring int)
>  * there's no int_big

This should be addressed yes, in this MR or in a subsequent one. If you want to add these new generators feel free to commit on my branch.

> So we end up having int_neg - but no int_pos - so we should rename int_pos_corners to nat_corners I guess?

Would it make sense to have `int_pos` which is just `let int_pos = nat` where we specify in the documentation to look for `nat_*` generators for more specific positive integers?

> Furthermore, it seems we are missing a generator of int_pos/nat_(small_)corners that utilizes int_pos_corners... 

It should also be addressed yes.

src/core/QCheck2.ml
vch9: I did not have the warning but I will double check. But I think the `gen` argument should prevent the warning yes.

src/core/QCheck2.ml
jmid: Sorry, brainfart on my end! 😅

vch9:
I really wish I can revive this PR, however, I'm still busy and I will be for a while. Once I get the time that'll be the first thing I'll do :)

PR #224, opened 2022-01-18
Add documentation for various functions

tmcgilchrist:
Add documentation for string_gen_of_size, small_printable_string, printable_char and printable.

### DIFF
--- a/src/core/QCheck.mli
+++ b/src/core/QCheck.mli
@@ -380,7 +380,7 @@ module Gen : sig
   val char : char t
   (** Generates characters upto character code 255. *)
 
-  val printable : char t (** Generates printable characters. *)
+  val printable : char t (** Generates printable ascii characters in the range 32 to 127 *)
 
   val numeral : char t (** Generates numeral characters. *)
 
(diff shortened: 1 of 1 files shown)


src/core/QCheck.mli
```diff
@@ -1203,6 +1203,7 @@ val numeral_char : char arbitrary
 (** Uniformly distributed over ['0'..'9']. *)
 
 val string_gen_of_size : int Gen.t -> char Gen.t -> string arbitrary
+(** Generates strings of exactly length [int Gen.t]. *)
```
jmid, line 1206: Thanks - and well spotted!

Since the size of the resulting strings vary with the output of the first argument (a "pure" Gen.t generator),
perhaps the documentation should be more along the lines of `Gen.string_size` that it calls underneath? 
```ocaml
  (** Builds a string generator from a (non-negative) size generator.
      Accepts an optional character generator (the default is {!char}). *)
```

tmcgilchrist: Agreed that's a better description.

src/core/QCheck.mli
```diff
@@ -1196,13 +1196,16 @@ val char : char arbitrary
...
+(** Uniformly distributed over a subset of printable ascii chars.
+    Ascii character codes 32 to 127.
+  *)
 
 val numeral_char : char arbitrary
 (** Uniformly distributed over ['0'..'9']. *)
 
 val string_gen_of_size : int Gen.t -> char Gen.t -> string arbitrary
+(** Builds a string generator from a (non-negative) size generator.
+    Accepts an optional character generator (the default is {!char}). *)
```
jmid, line 1208: Sorry - I just copy pasted the documentation for `Gen.string_size` for reference/convenience 🤷‍♂️
It is close but not entirely accurate as it takes an optional parameter which `string_gen_of_size` doesn't.

Can I propose:
```ocaml
(** Builds a string generator from a (non-negative) size generator and a character generator. *)
```
Then I promise not to boil more soup on this one... 😅

jmid:
Thank you for digging up the `printable_char` distribution and documenting it!
Could I ask you to add it to `Gen.printable` also? 🙏‍

tmcgilchrist:
@jmid Added a few more doc strings where I could. Please check that you're happy with them

PR #225, opened 2022-01-20
Clarify upper bound inclusion in int Gen functions

OlivierNicole:
I find the current document unclear about whether the upper bound is included or excluded for `Gen.int_bound` and `Gen.int_range`, especially since the behavior is not the same as `Random.int`. I propose to make very explicit.

### DIFF
--- a/src/core/QCheck.mli
+++ b/src/core/QCheck.mli
@@ -293,12 +293,13 @@ module Gen : sig
       @since 0.5.2 *)
 
   val int_bound : int -> int t
-  (** Uniform integer generator producing integers within [0... bound].
+  (** Uniform integer generator producing integers between [0] and [bound]
+      (inclusive).
       For [bound < 2^{30} - 1] uses [Random.State.int] for integer generation.
       @raise Invalid_argument if the argument is negative. *)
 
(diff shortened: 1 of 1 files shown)


jmid:
Thanks! 
Which versions of OCaml and qcheck are you running this on?
(potentially we could be observing a change of behavior in the new, underlying multicore RNG)

Here `int_bound` is inclusive:
```ocaml
utop # #require "qcheck";;
utop # open QCheck;;
utop # Gen.(generate ~n:20 (int_bound 5));;
- : int list = [0; 2; 3; 3; 2; 4; 4; 3; 3; 0; 2; 5; 3; 3; 5; 5; 3; 1; 0; 2]
utop # Gen.(generate ~n:20 (int_bound 5));;
- : int list = [5; 5; 5; 5; 5; 2; 2; 1; 5; 1; 4; 5; 2; 3; 2; 5; 2; 3; 5; 2]
```
Can I ask you try in ocaml/utop and check a couple of outputs?

OlivierNicole:
Sorry, you're right, my proposed change was wrong. The upper is not excluded (like with `Random.int`), but included! I pushed a fixed version.

PR #229, opened 2022-03-07
Fix wrong documentation occurrences of small_nat

### LINKED ISSUE #228: [Documentation] `QCheck.list` uses `Gen.nat`, not `Gen.small_nat` as suggested by documentation
The [documentation](https://c-cube.github.io/qcheck/0.18/qcheck-core/QCheck/#val-list) for the `list` arbitrary says the length of the resulting list is from `small_nat`. From what I can tell, the actual length is from `nat` (the arbitrary for small lists is `small_list`, where the documentation is correct).

`val list : 'a arbitrary -> 'a list arbitrary`
Generates lists with length generated by `small_nat`.

Should be

`val list : 'a arbitrary -> 'a list arbitrary`
Generates lists with length generated by `nat`.

jmid:
The documentation of some of the `arbitrary` generators is a bit off.
This fixes #228 - and some other ones I spotted:
- `string_gen`, `string`, `printable_string`, `numeral_string`, `list`, and `array` all use `Gen.nat` not `small_nat`
- it also turns the above occurrences into documentation links 
- a typo `if -> of` in `string_of_size`


vch9:
LGTM

PR #230, opened 2022-03-07
feat: add Gen.option

### LINKED ISSUE #226: Naming consistency: Gen.opt and Print.option
It seems to be more consistent to add `Gen.option` to match other modules.

PS: this is related to #162 but I believe the consistency between `Gen` and `Print` was not covered there.

favonia:
Close #226.


jmid:
LGTM!
- good point about naming consistency across Gen and Print
- I also appreciated that you added it for both QCheck.Gen and QCheck2.Gen

Should we update the deriver to use `option` while we are at it? @vch9 
https://github.com/c-cube/qcheck/blob/be9a0e78eadb663b5b2f9cb16e692ae93fb1cb68/src/ppx_deriving_qcheck/QCheck_generators.ml#L28

jmid:
Thanks for the update!

The CI is failing the tests. @vch9 wrote a bunch of good tests to ensure that the ppx produces the expected output.
These also need updating. You can run the tests locally with `make test` to catch such things early.

src/core/QCheck2.ml
```diff
@@ -355,13 +355,16 @@ module Gen = struct
...
-  (** [opt gen] shrinks towards [None] then towards shrinks of [gen]. *)
-  let opt ?(ratio : float = 0.85) (gen : 'a t) : 'a option t = fun st ->
+  (** [option gen] shrinks towards [None] then towards shrinks of [gen]. *)
+  let option ?(ratio : float = 0.85) (gen : 'a t) : 'a option t = fun st ->
     let p = RS.float st 1. in
     if p < (1. -. ratio)
     then Tree.pure None
     else Tree.opt (gen st)
 
+  (** [opt] is an alias of {!val:option} for backward compatibility. *)
```
vch9, line 365: Maybe we can begin the "deprecated" cycle here?

vch9:
Looks fine to me as well. I'd be in favor of deprecating the `opt` in favor of `option.

favonia:
I'm slightly worried that deprecating `Gen.opt` now would force people to change code with few benefits. Currently, `dune` by default would treat warnings as errors, which means CI or testing might suddenly fail due to the warning. I feel the deprecation cycle could begin after lots of function names have been changed due to #162 and other related issues, but not for just `Gen.opt`. That said, I'm happy to make the change if most people think it's worth it.

vch9:
> I'm slightly worried that deprecating `Gen.opt` now would force people to change code with few benefits. Currently, `dune` by default would treat warnings as errors, which means CI or testing might suddenly fail due to the warning. I feel the deprecation cycle could begin after lots of function names have been changed due to #162 and other related issues, but not for just `Gen.opt`. That said, I'm happy to make the change if most people think it's worth it.

I think the next release will (or should) include https://github.com/c-cube/qcheck/pull/223 which would introduce this deprecation cycle anyway. But as I'm not done with this PR the maintainers may publish a release without it, I could deprecate opt in https://github.com/c-cube/qcheck/pull/223 afterward, and then, leave your PR as is.

jmid:
Thanks for this - and sorry for the delay. Merging...

PR #234, opened 2022-04-02
Add unit and expect tests

jmid:
This is the second episode of nicking the ~~good~~ non-shrink-log parts from #172, #174, and #176.

Highlights:
- by renaming expect-files we can utilize `dune`'s expect test support and simplify the `dune` files
- tests of `check_exn`
- unit-tests for QCheck, corresponding to those from QCheck2
- a tree shrinker test for QCheck, corresponding to the one from QCheck2
- a bunch of tuple and bind-tests, both positive, negative, and with statistics

I see this last item as supplementing the `tup` tests from @vch9. I want to collect our tuple tests, but in order to keep the noise down I suggest we do so in a separate PR.

The PR is probably best read one commit after another. None of this should be controversial though.
Objections or a quick glance and thumbs-up would be appreciated, as I would like to get the last tests in play before they bit rot further...


jmid:
I'll merge these tests shortly unless I hear an outcry of objections... 😀

PR #237, opened 2022-04-16
More expect tests and decoupling expect test source and runner

jmid:
This is the third and last episode of extracting the non-shrink-log parts from #172, #174, and #176 thus getting the last tests from those in play before they bit rot further...

In addition to adding more expect tests, this PR also decouples the test source code from the actual runner.
This step paves the way for using a different runner - in particular #177 - which will be reduced to one commit with a new file.

Again: The PR is probably best read one commit after another. None of this should be controversial, so objections or a quick glance and thumbs-up would be appreciated.


c-cube:
very quick glance looked good :)

c-cube:
CI is broken because of https://github.com/ocaml/setup-ocaml/issues/479 I believe. I 'm still trying to figure it out on containers…

PR #244, opened 2022-05-30
Add `Test.make_neg`

jmid:
This PR adds a "primitive" `Test.make_neg` which lets one express negative property-based tests, that is, tests that are supposed to fail. This can be useful
- to safe guard the error-finding ability of a generator
- to supplement existing positive test (supposed to succeed)
- to document in test source code which tests are supposed to fail and succeed

The commercial Erlang QuickCheck has a similar operation, `fails`: http://quviq.com/documentation/eqc/eqc.html#fails-1

Here's an example:
```ocaml
open QCheck
let t1 = Test.make_neg ~name:"all ints are even" small_int (fun i -> i mod 2 = 0)
let t2 = Test.make ~name:"int double" small_int (fun i -> i + i = i * 2)
let t2' = Test.make_neg ~name:"int double" small_int (fun i -> i + i = i * 2)
let t3 = Test.make_neg ~name:"list rev concat" (pair (list small_int) (list small_int)) (fun (is,js) -> (List.rev is)@(List.rev js) = List.rev (is@js))
let t4 = Test.make ~name:"pos fail" small_int (fun i -> raise (Failure "pos-fail"))
let t5 = Test.make_neg ~name:"neg fail" small_int (fun i -> raise (Failure "neg-fail"))
;;
QCheck_runner.run_tests ~verbose:true [t1;t2;t2';t3;t4;t5];;
```
and corresponding output:
```
random seed: 442234180
generated error fail pass / total     time test name
[✓]    1    0    0    1 /  100     0.0s all ints are even
[✓]  100    0    0  100 /  100     0.0s int double
[✗]  100    0  100    0 /  100     0.0s int double
[✓]    1    0    0    1 /  100     0.0s list rev concat
[✗]    1    1    0    0 /  100     0.0s pos fail
[✗]    1    1    0    0 /  100     0.0s neg fail

--- Info -----------------------------------------------------------------------

Negative test all ints are even failed as expected (2 shrink steps):

3

Test int double failed:

Negative test int double succeeded but was expected to fail

--- Info -----------------------------------------------------------------------

Negative test list rev concat failed as expected (17 shrink steps):

([0], [1])

=== Error ======================================================================

Test pos fail errored on (5 shrink steps):

0

exception Failure("pos-fail")


=== Error ======================================================================

Test neg fail errored on (5 shrink steps):

0

exception Failure("neg-fail")

failure (1 tests failed, 2 tests errored, ran 6 tests)
- : int = 1
```

I have gone for keeping changes to a minimum and thus leave the test-generation-shrink loop and `Test.check_cell` untouched.
Negative tests are marked with a `bool` in their underlying `cell` - and it is instead the job of the runner to match this marker with `TestResult.t`s.

In the current design a test `Error` is considered a failure for both positive and negative tests (see pos-fail and neg-fail above).

Things missing:
- ~~internal tests of the new functionality~~
- ~~extend the alcotest runner~~
- ~~extend the ounit runner~~
- ~~add CHANGELOG entry~~

Input on the design and/or implementation are welcome.

### DIFF
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -1697,14 +1697,18 @@ module Test = struct
   let get_long_factor = QCheck2.Test.get_long_factor
 
   let make_cell ?if_assumptions_fail
-      ?count ?long_factor ?max_gen
+      ?count ?long_factor ?negative ?max_gen
   ?max_fail ?small:_removed_in_qcheck_2 ?retries ?name arb law
   =
   let {gen; shrink; print; collect; stats; _} = arb in
-  QCheck2.Test.make_cell_from_QCheck1 ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?retries ?name ~gen ?shrink ?print ?collect ~stats law
+  QCheck2.Test.make_cell_from_QCheck1 ?if_assumptions_fail ?count ?long_factor ?negative ?max_gen ?max_fail ?retries ?name ~gen ?shrink ?print ?collect ~stats law
 
   let make ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law =
     QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law)
+
+  let make_neg ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law =
+    let negative = Some true in
+    QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ?negative ?max_gen ?max_fail ?small ?retries ?name arb law)
 
   let fail_report = QCheck2.Test.fail_report
 
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -1703,12 +1703,11 @@ module Test = struct
   let {gen; shrink; print; collect; stats; _} = arb in
   QCheck2.Test.make_cell_from_QCheck1 ?if_assumptions_fail ?count ?long_factor ?negative ?max_gen ?max_fail ?retries ?name ~gen ?shrink ?print ?collect ~stats law
 
-  let make ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law =
-    QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law)
+  let make' ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name ~negative arb law =
+    QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name ~negative arb law)
 
-  let make_neg ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law =
-    let negative = Some true in
-    QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ?negative ?max_gen ?max_fail ?small ?retries ?name arb law)
+  let make = make' ~negative:false
+  let make_neg = make' ~negative:true
 
   let fail_report = QCheck2.Test.fail_report
 
--- a/src/core/QCheck2.mli
+++ b/src/core/QCheck2.mli
@@ -1587,10 +1587,10 @@ module Test : sig
   type 'a cell
   (** A single property test on a value of type ['a]. A {!Test.t} wraps a [cell]
       and hides its type parameter. *)
-     
+
   val make_cell :
     ?if_assumptions_fail:([`Fatal | `Warning] * float) ->
-    ?count:int -> ?long_factor:int -> ?max_gen:int -> ?max_fail:int -> ?retries:int ->
+    ?count:int -> ?long_factor:int ->  ?negative:bool -> ?max_gen:int -> ?max_fail:int -> ?retries:int ->
     ?name:string -> ?print:'a Print.t -> ?collect:('a -> string) -> ?stats:('a stat list) ->
      'a Gen.t -> ('a -> bool) ->
     'a cell
@@ -1601,6 +1601,7 @@ module Test : sig
         the test cases which satisfy preconditions.
       @param long_factor the factor by which to multiply count, max_gen and
         max_fail when running a long test (default: 1).
+      @param negative whether the test is expected to fail.
       @param max_gen maximum number of times the generation function
         is called in total to replace inputs that do not satisfy
         preconditions (should be >= count).
@@ -1621,11 +1622,11 @@ module Test : sig
 
   val make_cell_from_QCheck1 :
     ?if_assumptions_fail:([`Fatal | `Warning] * float) ->
-    ?count:int -> ?long_factor:int -> ?max_gen:int -> ?max_fail:int ->
+    ?count:int -> ?long_factor:int -> ?negative:bool -> ?max_gen:int -> ?max_fail:int ->
     ?retries:int -> ?name:string -> gen:(Random.State.t -> 'a) -> ?shrink:('a -> ('a -> unit) -> unit) ->
     ?print:('a -> string) -> ?collect:('a -> string) -> stats:'a stat list -> ('a -> bool) ->
     'a cell
-  (** ⚠️ Do not use, this is exposed for internal reasons only. ⚠️ 
+  (** ⚠️ Do not use, this is exposed for internal reasons only. ⚠️
 
       @deprecated Migrate to QCheck2 and use {!make_cell} instead.
    *)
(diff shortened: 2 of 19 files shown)


jmid:
Based on `QCheck2.Test.print_expected_failure` added in 8b1f246, commit 28d8891 updates QCheck_ounit to support `Test.neg_test`.

jmid:
I've completed initial `Test.make_neg` support in `QCheck_alcotest` in f932c28 and updated the CHANGELOG,
so this first version should be complete.

(I also discovered that I wrongly call it `Test.neg_test` is commit messages rather than `Test.make_neg`... :man_facepalming: :grimacing:)

jmid:
I've now
- added documentation of `Test.make_neg` (and polished the existing docs a bit)
- corrected the commit messages to use the right name binding `Test.make_neg` and force pushed the correction

Green light to merge @c-cube @vch9?

vch9:
I will have a look tomorrow morning, this seems to be a good idea!

src/core/QCheck.ml
```diff
@@ -1697,15 +1697,19 @@ module Test = struct
...
   let {gen; shrink; print; collect; stats; _} = arb in
-  QCheck2.Test.make_cell_from_QCheck1 ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?retries ?name ~gen ?shrink ?print ?collect ~stats law
+  QCheck2.Test.make_cell_from_QCheck1 ?if_assumptions_fail ?count ?long_factor ?negative ?max_gen ?max_fail ?retries ?name ~gen ?shrink ?print ?collect ~stats law
 
   let make ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law =
     QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law)
 
+  let make_neg ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law =
+    let negative = Some true in
+    QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ?negative ?max_gen ?max_fail ?small ?retries ?name arb law)
```
vch9, line 1711:
```suggestion
    let negative = true in
    QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ~negative ?max_gen ?max_fail ?small ?retries ?name arb law)
```

Also that could be a good idea to have `make'` having all these parameters and then having:
```ocaml
let make = make'

let make_neg = make' ~negative:true
```

src/core/QCheck2.mli
```diff
@@ -1660,6 +1669,18 @@ module Test : sig
...
   *)
 
+  val make_neg :
+    ?if_assumptions_fail:([`Fatal | `Warning] * float) ->
+    ?count:int -> ?long_factor:int -> ?max_gen:int -> ?max_fail:int -> ?retries:int ->
+    ?name:string -> ?print:('a Print.t) -> ?collect:('a -> string) -> ?stats:('a stat list) ->
+    'a Gen.t -> ('a -> bool) -> t
+  (** [make_neg gen prop] builds a test that checks property [prop] on instances
+      of the generator [gen].
+      The test is considered negative, meaning that it is expected to fail.
```
vch9, line 1679: "expected to fail" makes me think of actual failures. Would something like "does not meet the property" be better? I'm not sure, you can ignore this comment

src/runner/QCheck_base_runner.ml
```diff
@@ -406,17 +421,24 @@ let run_tests
...
         (total + 1, fail + 1, error, warns)
-      | R.Failed_other {msg} ->
+      | R.Failed {instances=l}, false ->
+        if verbose then List.iter (print_expected_failure ~colors out cell) l;
+        (total + 1, fail, error, warns)
+      | R.Failed_other {msg}, _ ->  (* Failed_other is also considered a failure *)
         print_fail_other ~colors out cell msg;
         (total + 1, fail + 1, error, warns)
-      | R.Error {instance=c_ex; exn; backtrace=bt} ->
+      | R.Error {instance=c_ex; exn; backtrace=bt}, _ -> (* Error is always considered a failure *)
```
vch9, line 441: That's a good thing. It's the first thing I checked when I opened your mr :+1:

vch9:
Looks goo to merge modulo my very minor remarks!

src/core/QCheck2.mli
jmid: I think this is a relevant concern to help distinguish a "failed property" from a "failed test".
In 9199574 I've changed the formulation to "expected not to satisfy the tested property" in both the QCheck and QCheck2 interfaces to help clarify.

vch9: It's better yes, thank you. Could you also modify the changelog with your modification?

jmid: Fixed in daa57f8

src/core/QCheck.ml
vch9: What do you think about:
```diff
diff --git a/src/core/QCheck.ml b/src/core/QCheck.ml
index 99a05e7..04f2137 100644
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -1703,12 +1703,12 @@ module Test = struct
   let {gen; shrink; print; collect; stats; _} = arb in
   QCheck2.Test.make_cell_from_QCheck1 ?if_assumptions_fail ?count ?long_factor ?negative ?max_gen ?max_fail ?retries ?name ~gen ?shrink ?print ?collect ~stats law
 
-  let make ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law =
-    QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law)
+  let make' ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name ?negative arb law =
+    QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name ?negative arb law)
 
-  let make_neg ?if_assumptions_fail ?count ?long_factor ?max_gen ?max_fail ?small ?retries ?name arb law =
-    let negative = true in
-    QCheck2.Test.Test (make_cell ?if_assumptions_fail ?count ?long_factor ~negative ?max_gen ?max_fail ?small ?retries ?name arb law)
+  let make = make' ~negative:false
+
+  let make_neg = make' ~negative:true
```

jmid: Ah, thanks!
Sorry for the brain fart. I tried something along these lines and got myself more confused by the resulting type error :sweat_smile: 

I went with making `negative` a labeled (mandatory) parameter in e068041 since there are only two clients (`make` and `make_neg`) to this internal helper function - and both of them pass the parameter explcitly.

vch9:
I can not close the threads (I don't know if it's a permission issue or a github thing though), but, I think they were all treated.

jmid:
Thanks @vch9 - I'll squash a bit and then merge.

alopezz:
Was it intended for `Test.check_exn` to ignore the "polarity" of the test as per this feature?

I was just trying out the library and after seeing [this](https://c-cube.github.io/qcheck/0.20/qcheck-core/QCheck/index.html#tests):

> - [Test.make](https://c-cube.github.io/qcheck/0.20/qcheck-core/QCheck/Test/index.html#val-make) to build a test,
>  - [Test.make_neg](https://c-cube.github.io/qcheck/0.20/qcheck-core/QCheck/Test/index.html#val-make_neg) to build a negative test that is expected not to satisfy the tested property,
>   - [Test.check_exn](https://c-cube.github.io/qcheck/0.20/qcheck-core/QCheck/Test/index.html#val-check_exn) to run a single test with a simple runner.

I started to try with some example of my own and I saw that `Test.check_exn` was doing the same for `make` and `make_neg` when using the same property.

E.g. I would have expected the following to raise an exception but it doesn't, which is confusing.

```ocaml
let test = Test.make_neg ~name:"int double" small_int (fun i -> i + i = i * 2) in Test.check_exn test
```

From a quick at the look it sounds like it might be possible to factor out the individual implementations for ounit and alcotest by moving those somewhere under `check_cell_exn` since they both rely on it, which might provide a more consistent behavior. If that sounds good I wouldn't mind trying to work on the change myself.

jmid:
First off, I agree that this is confusing - and the lack of documentation is certainly not helping here.
In `QCheck.Test` the relevant functions are implemented by calling the equivalent `QCheck2.Test` ones, which are slightly better documented. For example, https://c-cube.github.io/qcheck/0.20/qcheck-core/QCheck2/Test/index.html#val-check_cell
```
check_cell ~long ~rand test generates up to count random values of type 'a using Gen.t and the random state st. The predicate law is called on them and if it returns false or raises an exception then we have a counter-example for the law.
```

When I implemented the first `make_neg` version, it just seemed natural to build on `Test.check_exn` as a primitive.
It furthermore had the advantage of not having to change the existing public signatures.
I'm unsure what would be a reasonable interface for a `unit`-returning `check_exn` function in the case of success or failure of negative tests to communicate both "negative test did not fail as expected" and "negative test failed as expected and this is the counterexample".
Another aspect is making such a change in a greater context. As I see it, there are really 3 low-level drivers `check_cell`, `check_cell_exn` and `check_exn`. Here it is surprising, but at least consistent that all three ignore the test polarity... :sweat_smile: 

The simple answer is probably to document the current behaviour better.

alopezz:
@jmid Thanks for your reply!

I made a small PR (#271) to better illustrate what I meant. I'm not sure that `check_exn` would necessarily need to provide a counterexample in the "negative test failed as expected" case. Let me know what you think!

PR #242, opened 2022-04-21
Shrinker improvements to QCheck list, string, and function

jmid:
This PR contains the QCheck(1) shrinker improvements to `list`, `string`, and functions described in #235.
The essential part extracted from https://github.com/c-cube/qcheck/pull/235#issuecomment-1103781517:

> Executive summary:
> - the improved list, string, and function shrinkers collectively reduce the runtime of the benchmark from 78.402s to 5.456s on my laptop

The reduced shrink counts in, e.g., test/core/QCheck_expect_test.expected are nice, but beware that this only counts **successful shrink attempts**. https://github.com/c-cube/qcheck/pull/235#issuecomment-1103781517 documents how the PR reduces the **total shrink attempts** significantly, e.g.,
- `string never has a \255 char` is reduced from 9365 attempts to 38 (on seed 6789)
- `lists shorter than 432` is reduced from 5118102 attempts to 19461 (on seed 1234)
- `fold_left fold_right uncurried` is reduced from 80630 attempts to 199 (on seed 1234)

I would be grateful if someone could give this a quick spin with an `opam pin` of https://github.com/jmid/qcheck/tree/shrinker-improvements-list-string-fun on their regular QCheck tests before merging.
- @c-cube perhaps on containers?
- @tmcgilchrist perhaps on some of your CI work?

PR #235, opened 2022-04-03
Shrinker improvements

jmid:
This PR combines QCheck(1) shrinker improvements to `list`, `string`, functions, and `int`.
The improvements arose from studying the benchmark outputs from #177 across a range of equivalent tests.
Here it became clear that both the QCheck(1) and QCheck2 shrinkers each had some sore spots.
As a first step, I set out to get QCheck(1) up to speed with QCheck2's performance on the tests where its shrinkers were lacking behind.

I don't expect to merge the PR in the current form, partly because it includes an `int` shrinker along the lines of the somewhat controversial #173, but
- I wanted to convey the bigger picture
- I was curious to hear feedback from potential users.

@vch9 : if you (or others) still have QCheck(1) tests (needing shrinking) I'd be curious to hear if you observe a difference
when pinning https://github.com/jmid/qcheck/tree/shrinker-improvements

Whereas these improvements have been obtained over artificial benchmark programs, I've been bitten in the past by shrinker performance in bigger developments, e.g.,
- in the [shrinker]( https://github.com/jmid/wasm-prop-tester/blob/master/src/shrinker.ml) for our Wasm generator where it made a difference to reduce integers to zero quickly (in the common case most integer values are irrelevant).
- in the shrinker for [multicoretests](https://github.com/jmid/multicoretests), where the non-deterministic behavior is extra challenging. Here reproducing a failing run O(n log n) times has a much lower probability than O(log n) times, which means that a algorithmically faster shrinker performing fewer, well-chosen reduction attempts will produce smaller counterexamples than a "more exhaustive" one.

The core is an improved `list_spine` shrinker. As an added bonus this implicitly improves function shrinking.
Secondly, we can reuse the improved `list` shrinker for `string`s - by breaking them up in `char list`s - and then use the `char`-shrinker to reduce the individual `char`s.  Overall, these three were identified as QCheck(1) sore spots in #177.

Despite the reduced complexity we are able to arrive at the same counterexamples - or simpler(!) as should
be clear from the expect-test logs.

Here's the new and simpler shrinker of list spines using recursion:
```ocaml
  let rec list_spine l yield =
    let rec split l len acc = match len,l with
      | _,[]
      | 0,_ -> List.rev acc, l
      | _,x::xs -> split xs (len-1) (x::acc) in
    match l with
    | [] -> ()
    | [_] -> yield []
    | [x;y] -> yield []; yield [x]; yield [y]
    | _::_ ->
      let len = List.length l in
[...]

PR #245, opened 2022-06-08
add bytes generator, printer and shrinker

### LINKED ISSUE #83: Add missing 'bytes' generators/shrinkers/printers/...
We are missing generators/shrinkers/printers/... for the `bytes` type.

n-osborne:
Add the missing primitives for `bytes`

Close #83


vch9:
Cool! Haven't look very deep in the code yet but it could be nice to both add the generators to `QCheck2` and append the changelog :).

src/core/QCheck.mli
```diff
@@ -397,6 +397,26 @@ module Gen : sig
...
+  (** Builds a bytes generator. Bytes size is generated by {!nat}.
+      Accepts an optional character generator (the default is {!char}).
+      See also {!bytes_of} and {!bytes_readable} for versions without
+      optional parameters. *)
+
+  val bytes_of : char t -> bytes t
+  (** Builds a bytes generator using the given character generator. *)
+
+  val bytes_printable : bytes t
+  (** Builds a bytes generator using the {!printable} character generator. *)
```
vch9, line 414:
```suggestion
  (** Generator using the {!printable} character generator. *)
```

src/core/QCheck.mli
```diff
@@ -397,6 +397,26 @@ module Gen : sig
...
+
+  val small_bytes : ?gen:char t -> bytes t
+  (** Builds a bytes generator, length is {!small_nat}
+      Accepts an optional character generator (the default is {!char}). *)
```
vch9, line 418: Should we add the tag `@since NEXT_RELEASE` as well as updating the changelog for the new generators?

src/core/QCheck.ml
```diff
@@ -86,15 +86,20 @@ let _opt_sum a b = match a, b with
 
 let sum_int = List.fold_left (+) 0
 
-(* Included for backwards compatibility, pre 4.13 *)
-let string_fold_right f s acc =
-  let len = String.length s in
+let _fold_righ length get f x acc =
```
vch9, line 89:
```suggestion
let _fold_rigth length get f x acc =
```

jmid: eehh, I believe it is spelled `right`.
Now allow me to include misspell myslef - casting the first stone and all... :wink:

jmid:
This is great indeed, thanks @n-osborne! :smiley: 
I second @vch9's suggestion of also adding the bytes combinators to QCheck2 to keep the two versions aligned.

I would also appreciate it if you would include tests of them in `test/core/QCheck_tests.ml` and `test/core/QCheck2_tests.ml`.
You can run these with `make tests`.
There are already `string` tests in `Generator`, `Shrink`, and `Statistics` which could be mirrored for `bytes`.

src/core/QCheck.ml
n-osborne: Thanks for spotting this typo!

src/core/QCheck2.mli
```diff
@@ -289,6 +289,47 @@ module Gen : sig
...
+
+  val bytes_of : char t -> bytes t
+  (** Builds a bytes generator using the given character generator.
+
+      Shrinks on the number of characters first, then on the characters.
+
+      @since NEXT_RELEASE *)
+
+  val bytes_printable : bytes t
+  (** Builds a bytes generator using the {!printable} character generator.
```
vch9, line 318:
```suggestion
  (** Generator using the {!printable} character generator.
```

src/core/QCheck2.mli
```diff
@@ -289,6 +289,47 @@ module Gen : sig
...
+  val bytes_size : ?gen:char t -> int t -> bytes t
+  (** Builds a bytes generator from a (non-negative) size generator.
+      Accepts an optional character generator (the default is {!char}).
+
+      Shrinks on the number of characters first, then on the characters.
+
+      @since NEXT_RELEASE *)
+
+  val bytes : bytes t
+  (** Builds a bytes generator. Bytes size is generated by {!nat}.
```
vch9, line 301:
```suggestion
  (** Bytes generator. Bytes size is generated by {!nat}.
```

src/core/QCheck2.mli
```diff
@@ -289,6 +289,47 @@ module Gen : sig
...
+      @since NEXT_RELEASE *)
+
+  val bytes_printable : bytes t
+  (** Builds a bytes generator using the {!printable} character generator.
+
+      Shrinks on the number of characters first, then on the characters.
+
+      @since NEXT_RELEASE *)
+
+  val small_bytes : ?gen:char t -> bytes t
```
vch9, line 324: Note that `?gen` can not be omitted with the abstraction of generators in QCheck2

test/core/QCheck2_tests.ml
```diff
@@ -727,6 +762,16 @@ module Stats = struct
...
     Test.make ~name:"char code dist" ~count:500_000 ~stats:[("char code", Char.code)] Gen.char (fun _ -> true)
 
+  let bytes_len_tests =
+    let len = ("len",Bytes.length) in
+    [
+      Test.make ~name:"bytes_size len dist"      ~count:5_000 ~stats:[len] Gen.(bytes_size (int_range 5 10)) (fun _ -> true);
+      Test.make ~name:"bytes len dist"           ~count:5_000 ~stats:[len] Gen.bytes                         (fun _ -> true);
+      Test.make ~name:"bytes_of len dist"        ~count:5_000 ~stats:[len] Gen.(bytes_of (return 'a'))       (fun _ -> true);
+      Test.make ~name:"bytes_printable len dist" ~count:5_000 ~stats:[len] Gen.bytes_printable               (fun _ -> true);
+      Test.make ~name:"small_bytes len dist"     ~count:5_000 ~stats:[len] Gen.(small_bytes ~gen:char)(*ugh*)(fun _ -> true);
```
vch9, line 772: Is there any reason for the `(*ugh*)`

vch9:
Looks good!

src/core/QCheck2.mli
n-osborne: But `string_size` line 333 and `small_string` line 363 ha an optional `?gen` argument.
Do I remove them too?

test/core/QCheck2_tests.ml
n-osborne: Not really apart from copy-pasting the `string` tests.
As, ``small_string` line 782 has this `(* ugh *)` too, I think this is related to the presence of the optional `gen` argument.

src/core/QCheck2.mli
n-osborne: `string_size` line 333 and `small_string` line 363 also have an optional `?gen` argument.
Do I remove them too?

test/core/QCheck2_tests.ml
n-osborne: Not really apart from copy-pasting the `string` tests.
As, ``small_string` line 782 has this `(* ugh *)` too, I think this is related to the presence of the optional `gen` argument.

jmid: I am to blame for the `(*ugh*)` - and if I remember correctly(tm) the reason was precisely being forced to provide an "optional `char`".

n-osborne: I've removed them, and fixed what caused them.

jmid: The problem is that for 1-argument combinators, the type-checker gets confused if the single argument is optional.
In effect, the "optional argument" is actually not optional - but required (for the code to type check).
This is a known issue with some of the existing combinators (see, e.g., #161).

Note: Since this is not a problem for 2-,3-,...-argument combinators such as `bytes_size` that also take a regular argument - I would favor to keep its signature aligned with `string_size` (still taking an optional parameter). Otherwise, it would be a breaking API change, AFAICS.

For `Gen.small_bytes` (and `Gen.small_string`) one option would be to
- keep the argument to both but make it labelled and hence mandatory (like in #161)
- add combinators
  - `Gen.bytes_small : bytes t`
  - `Gen.bytes_small_of : char t -> bytes t`
- (and similarly for `Gen.small_string`)

This has the advantage of
- not breaking backward compatibility
- aligning generators better, so that everything `string` or `bytes`-related start with the prefix `string...` or `bytes...` - which make them easier to search for using IDE-feedback...

We should agree on such a design first though. Thoughts @vch9 @c-cube ?

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changes
 
+## NEXT_RELEASE
```
jmid, line 3: We still haven't released 0.19 so you can just include these below (it is building up to quite a release...)

src/core/QCheck.ml
```diff
@@ -763,16 +779,19 @@ module Shrink = struct
...
-      (string_fold_right (fun c acc -> c::acc) s [])
+      (fold (fun c acc -> c::acc) s [])
       (fun cs ->
          List.iter (fun c -> Buffer.add_char buf c) cs;
-         let s = Buffer.contents buf in
+         let s = get buf in
          Buffer.clear buf;
          yield s)
 
+  let bytes = _shrink_string_and_bytes bytes_fold_right (fun b -> Buffer.contents b |> Bytes.of_string)
```
jmid, line 792: There's `Buffer.to_bytes` since 4.02 which I think can simplify the above to:
```ocaml
let bytes = _shrink_string_and_bytes bytes_fold_right Bytes.to_bytes
```

test/core/QCheck2_tests.ml
n-osborne: I am happy to put back the `~gen`. Backward compatibility is a big argument.

src/core/QCheck.ml
```diff
@@ -763,16 +779,19 @@ module Shrink = struct
...
+      (fold (fun c acc -> c::acc) s [])
       (fun cs ->
          List.iter (fun c -> Buffer.add_char buf c) cs;
-         let s = Buffer.contents buf in
+         let s = get buf in
          Buffer.clear buf;
          yield s)
 
+  let bytes = _shrink_string_and_bytes bytes_fold_right Buffer.to_bytes
+  let string = _shrink_string_and_bytes string_fold_right Buffer.contents
```
jmid, line 793: This touches the `string` shrinker and hence triggers an alarm bell in me :sweat_smile: 

The basic idea of the current `string` shrinker is
- to convert a `string` (or `bytes`) to a `list`
- fall back on the `list` shrinker
- map each shrunk `list` result back to a `string` (or `bytes`) using a `Buffer`.
    
1. Abstracting over the involved fold/get operations is certainly a valid approach to avoid code duplication. However I'm wondering whether it makes the shrinkers needlessly complex/higher-order :thinking:  
   Here's another way to obtain something similar without copy-pasting, which I think is clearer:
   Keeping the current `string` definition, define `bytes` as 
   ```ocaml
   let bytes b = Iter.map Bytes.of_string (string (Bytes.to_string b))
   ```
   By not touching the `string` shrinker we are sure not to affect its performance. Since `string` is more common than `bytes` this also optimizes for the common case.
   Alternatively, one could also keep an equivalent `bytes` version and define `string` in terms of it - perhaps using `Bytes.unsafe_to_string` to save copying during conversion (would happen for each shrinking candidate). 

2. I spent quite some time performance tweaking the shrinkers in #242 and #235 - even for the `string` shrinker (see https://github.com/c-cube/qcheck/pull/242#discussion_r867250149). I would thus be sad to see some of that hard-earned performance lost because, e.g., the compiler was unable to inline some of these function arguments. If we end up going with the `Shrink.string` changes I would therefore be grateful if you confirm with benchmarks that my concerns are unwarranted. Here are instructions for the current shrinker benchmarks:
   ```
   dune build test/core/shrink_benchmark.exe
   _build/default/test/core/shrink_benchmark.exe
   # shrinker benchmark output is now available in shrink_bench.log
   ```

src/core/QCheck.ml
```diff
@@ -946,9 +966,10 @@ module Observable = struct
     type 'a t = 'a -> 'a -> bool
 
     let int : int t = (=)
+    let bytes : bytes t = (=)
     let string : string t = (=)
     let bool : bool t = (=)
-    let float : float t = (=)
+    let float : float t = Float.equal
```
jmid, line 972: I think it would be cleaner to just submit the two small `float` equality fixes as a separate PR.
It will make it easier to track that fix when decoupled from this larger PR.

src/core/QCheck2.mli
jmid: Removing the tilde `~` should make the argument labeled in the signature:
```ocaml
  val small_bytes : gen:char t -> bytes t
```
(the same should probably be done for `small_string`)

src/core/QCheck2.ml
```diff
@@ -698,6 +698,14 @@ module Gen = struct
   let string_size ?(gen = char) (size : int t) : string t =
     bytes_size ~gen size >|= Bytes.unsafe_to_string
 
+  let bytes : bytes t = bytes_size nat
+
+  let bytes_of gen = bytes_size ~gen nat
+
+  let bytes_printable = bytes_size ~gen:printable nat
+
+  let small_bytes ?gen st = bytes_size ?gen small_nat st
```
jmid, line 707: The definition should probably also be updated to be labeled:
```ocaml
let small_bytes ~gen st = ....
```
(and similarly for `small_string`)

src/core/QCheck.ml
n-osborne: That makes sense :+1:

src/core/QCheck.ml
```diff
@@ -92,7 +92,7 @@ let string_fold_right f s acc =
   let rec loop i acc =
     if i<0
     then acc
-    else loop (i-1) (f s.[i] acc) in
+    else loop (i-1) (f (String.get s i) acc) in
```
jmid, line 95: I would probably just revert this change to keep the diff minimal.

src/core/QCheck.ml
```diff
@@ -946,9 +960,10 @@ module Observable = struct
     type 'a t = 'a -> 'a -> bool
 
     let int : int t = (=)
+    let bytes : bytes t = (=)
     let string : string t = (=)
     let bool : bool t = (=)
-    let float : float t = (=)
+    let float : float t =  (=)
```
jmid, line 966: Nit: inserted space

src/core/QCheck2.ml
```diff
@@ -972,11 +985,13 @@ module Observable = struct
...
     let int : int t = (=)
 
+    let bytes : bytes t = (=)
+
     let string : string t = (=)
 
     let bool : bool t = (=)
 
-    let float : float t = (=)
+    let float = (=)
```
jmid, line 994: I think the previous type annotation was there to allow the compiler to specialize the polymorphic equality into a (faster) float equality.

With the float equality fix coming in another PR, I would again prefer to leave this line untouched to keep the diff minimal.

src/core/QCheck2.mli
```diff
@@ -1267,7 +1319,7 @@ module Observable : sig
   (** [quad o1 o2 o3 o4] is an observable of quadruples of [('a * 'b * 'c * 'd)]. *)
 end
 
-  
+
```
jmid, line 1322: Nit: here follows 3 whitespace changes making the diff bigger than necessary

src/core/QCheck.mli
```diff
@@ -397,6 +397,31 @@ module Gen : sig
...
+
+  val bytes_of : char t -> bytes t
+  (** Builds a bytes generator using the given character generator.
+      @since NEXT_RELEASE *)
+
+  val bytes_printable : bytes t
+  (** Generator using the {!printable} character generator.
+      @since NEXT_RELEASE *)
+
+  val small_bytes : ?gen:char t -> bytes t
```
jmid, line 420: There's now `QCheck.bytes_small : bytes arbitrary` and `QCheck2.Gen.bytes_small` (yay!)
but no `QCheck.Gen.bytes_small`?

src/core/QCheck2.ml
n-osborne: Do I remove the type annotation in `QCheck.ml` too?

jmid: To keep this PR as focused as possible (topic: `bytes` combinators - but also touching `string`), I would prefer "no diff" on any line involving `float`.

That is not to say that a `float` equality fix is not welcome! :pray: It just helps keeping PRs small, managable, and reviewable.
I would expect your forthcoming `float` fix PR to be a two-line diff (one change in QCheck.ml + one in QCheck2.ml) to be accepted and merged immediately.

jmid:
Thanks @n-osborne! :pray:  I'm sorry that this drifted into a design PR as this takes extra time to get right.

To summarize, with this PR we arrive at the below `bytes` and `string` interfaces:

### `QCheck.Gen`:
```ocaml
  val bytes_size : ?gen:char t -> int t -> bytes t
  val bytes : ?gen:char t -> bytes t
  val bytes_of : char t -> bytes t
  val bytes_printable : bytes t
  val bytes_small : ?gen:char t -> bytes t
  val string : ?gen:char t -> string t
  val string_of : char t -> string t
  val string_readable : string t  (* deprecated *)
  val string_printable : string t
```
Here, we could use this opportunity to offer consistent, non-optional `small` combinators for both `bytes` and `string`:
```ocaml
  val bytes_small : bytes t
  val bytes_small_of : char t -> bytes t
  val string_small : string t
  val string_small_of : char t -> string t
```
(the last two don't necessarily need to be part of this PR)

### `QCheck`:
```ocaml
  val bytes_gen_of_size : int Gen.t -> char Gen.t -> bytes arbitrary
  val bytes_gen : char Gen.t -> bytes arbitrary
  val bytes : bytes arbitrary
  val bytes_small : bytes arbitrary
  val bytes_of_size : int Gen.t -> bytes arbitrary
  val bytes_printable : bytes arbitrary
  val string_gen_of_size : int Gen.t -> char Gen.t -> string arbitrary
  val string_gen : char Gen.t -> string arbitrary
  val string : string arbitrary
  val small_string : string arbitrary
  val string_of_size : int Gen.t -> string arbitrary
  val printable_string : string arbitrary
  val printable_string_of_size : int Gen.t -> string arbitrary
  val small_printable_string : string arbitrary
  val numeral_string : string arbitrary
  val numeral_string_of_size : int Gen.t -> string arbitrary
```
Here we may want to add `string_small`, `string_printable`, ... in a move toward offering type-prefixed names - again not necessarily in this PR.

### `QCheck2.Gen`:
```ocaml
  val bytes_size : ?gen:char t -> int t -> bytes t
  val bytes : bytes t
  val bytes_of : char t -> bytes t
  val bytes_printable : bytes t
  val bytes_small : gen:char t -> bytes t
  val string : string t
  val string_of : char t -> string t
  val string_printable : string t
  val string_small : gen:char t -> string t
  val small_string : gen:char t -> string t
```
Looking at it now, I think labeling the parameter to `small_string` is the right backwards-compatible fix.
However, we might consider aligning the 1-argument `bytes*` and `string*` combinators as non-labeled akin to `QCheck.Gen` above:
```ocaml
  val bytes_small : bytes t
  val bytes_small_of : char t -> bytes t
  val string_small : string t
  val string_small_of : char t -> string t
```

I would appreciate your eyes and input on these interfaces @vch9 and @c-cube

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changes
 
+## 0.19
```
jmid, line 3: I propose omitting this line or changing it to `0.19.2` or `NEXT_RELEASE`

src/core/QCheck.ml
```diff
@@ -1118,14 +1145,20 @@ let string_gen gen =
...
 let string_of_size size = string_gen_of_size size Gen.char
-let small_string = string_gen_of_size Gen.small_nat Gen.char
-
-let printable_string = string_gen Gen.printable
-let printable_string_of_size size = string_gen_of_size size Gen.printable
-let small_printable_string = string_gen_of_size Gen.small_nat Gen.printable
-
-let numeral_string = string_gen Gen.numeral
-let numeral_string_of_size size = string_gen_of_size size Gen.numeral
+let string_small = string_gen_of_size Gen.small_nat Gen.char
```
jmid, line 1148: AFAIU, this adds new `arbitrary` `string` aliases without exposing them in the interface `QCheck.mli`:
- `string_small`
- `string_printable`
- `string_printable_of_size`
- `string_small_printable`
- `string_numeral`
- `string_numeral_of_size`

I think they should also be added to the CHANGELOG.

jmid (changes requested):
Thanks! I've now taken another pass over this.

I spotted that some of these new `string arbitrary` generator aliases are not exposed in the interface. Ideally this would have been caught by just writing a simple test - but on the other hand, we shouldn't have to write duplicate tests for both aliases... :shrug:  

(I also spotted that some of the shrinkers for the above are broken but that is not caused by this PR, so I'll file a separate issue).

src/core/QCheck.ml
n-osborne: It is added, thanks for spotting that :smile:.

CHANGELOG.md
n-osborne: Done

src/core/QCheck.mli
```diff
@@ -1254,31 +1302,67 @@ val string : string arbitrary
 val small_string : string arbitrary
 (** Same as {!string} but with a small length (ie {!Gen.small_nat} ). *)
 
+val string_small : string arbitrary
+(** Same as {!string} but with a small length (ie {!Gen.small_nat} ).
+    @since NEXT_RELEASE *)
+
```
jmid, line 1308: I propose documenting that this is a convenient alias for `small_string` in the interface, and similarly for the other `string arbitrary` generator aliases. Otherwise it is hard as a new-comer to tell the difference between two similarly named generators with identical documentation :shrug:

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## 0.20
 
+- add `{QCheck,QCheck2}.Gen.{bytes_size,bytes,bytes_of,bytes_printable,small_bytes}`
```
jmid, line 5: Nit: the last name if off. Should be 
```
- add `{QCheck,QCheck2}.Gen.{bytes_size,bytes,bytes_of,bytes_printable,bytes_small}`
```

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
+- add `QCheck.{Print,Shrink,Observable}.bytes`
+- add `QCheck2.{Print,Shrink}.bytes`
+- add `QCheck.{bytes_gen_of_size,bytes_gen,bytes,small_bytes,bytes_of_size,printable_bytes,printable_bytes_of_size,numeral_bytes,numeral_bytes_of_size}`
```
jmid, line 8: These names are also off.
AFAICS, this adds 6 `arbitrary` `bytes` combinators - all prefixed with `bytes_` - so the last four should be scrapped and `small_bytes` renamed.

CHANGELOG.md
jmid: Wait! Some of these additions are also listed below in line 9 :thinking: 
```
- add `{QCheck,QCheck2}.Gen.{bytes_size,bytes,bytes_of,bytes_printable,bytes_small,string_small}`
```

CHANGELOG.md
jmid: Some of these are also repeated in line 12:
```
- add `QCheck.{bytes_gen_of_size,bytes_gen,bytes,bytes_small,bytes_of_size,bytes_printable}`
```

src/core/QCheck2.ml
```diff
@@ -979,11 +992,13 @@ module Observable = struct
...
     let int : int t = (=)
 
+    let bytes : bytes t = (=)
+
     let string : string t = (=)
 
     let bool : bool t = (=)
 
-    let float = Float.equal
+    let float : float t = (=)
```
jmid, line 1001: This change reverts a fix from #249

jmid:
Thanks for the `printable` `bytes` shrinker fix!

Taking another pass of it I noticed that the CHANGELOG entries are off - and there is also an unintentional reversion of #249 resulting from all the rebasing.

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
+- add `{QCheck,QCheck2}.Gen.{bytes_size,bytes,bytes_of,bytes_printable,bytes_small}`
+- add `QCheck.{Print,Shrink,Observable}.bytes`
+- add `QCheck2.{Print,Shrink}.bytes`
+- add `QCheck.{bytes_gen_of_size,bytes_gen,bytes,bytes_small,bytes_of_size,bytes_printable}`
+- add `{QCheck,QCheck2}.Gen.{bytes_size,bytes,bytes_of,bytes_printable,bytes_small,string_small}`
+- add `QCheck.{Print,Shrink,Observable}.bytes`
+- add `QCheck2.{Print,Shrink}.bytes`
+- add `QCheck.{bytes_gen_of_size,bytes_gen,bytes,bytes_small,bytes_of_size,bytes_printable}`
+- add `QCheck.{string_small,string_printable,string_printable_of_size,string_small_printable,string_numeral,string_numeral_of_size}`
+- `QCheck2.{small_string}` character generator argument is no more optional
```
jmid, line 14: This still contains redundancy between the individual lines.
I propose reducing it to:
```
- add `{QCheck,QCheck2}.Gen.{bytes_size,bytes,bytes_of,bytes_printable,bytes_small}`
- add `QCheck2.Gen.string_small`
- add `QCheck.{Print,Shrink,Observable}.bytes`
- add `QCheck2.{Print,Shrink}.bytes`
- add `QCheck.{bytes_gen_of_size,bytes_gen,bytes,bytes_small,bytes_of_size,bytes_printable}`
- add `QCheck.{string_small,string_printable,string_printable_of_size,string_small_printable,string_numeral,string_numeral_of_size}`
- `QCheck2.small_string` character generator argument is no more optional
```

jmid (changes requested):
Thanks. 
The changelog is still not quite right. I've added a proposal.

jmid (approved):

jmid:
I've committed a few more changes I proposed in https://github.com/c-cube/qcheck/pull/245#issuecomment-1162975158

For reference, below follows a summary of the resulting `bytes` and `string` combinator interfaces.
Ignoring the `(* to-be-deprecated: *)` sections, I believe this represents a good step towards more consistent `bytes` and `string` combinator names.


`QCheck.Gen`
--------------------
```ocaml
val bytes_size : ?gen:char t -> int t -> bytes t
val bytes : ?gen:char t -> bytes t
val bytes_of : char t -> bytes t
val bytes_small : bytes t
val bytes_small_of : char t -> bytes t
val bytes_printable : bytes t

val string : ?gen:char t -> string t
val string_of : char t -> string t
val string_small : string t
val string_small_of : char t -> string t
val string_printable : string t

(* deprecated or to-be-deprecated: *)
val string_readable : string t
```

`QCheck` `arbitrary`
---------------------------
```ocaml
val bytes_gen_of_size : int Gen.t -> char Gen.t -> bytes arbitrary
val bytes : bytes arbitrary
val bytes_of : char Gen.t -> bytes arbitrary
val bytes_small : bytes arbitrary
val bytes_small_of : char Gen.t -> bytes arbitrary
val bytes_of_size : int Gen.t -> bytes arbitrary
val bytes_printable : bytes arbitrary

val string_gen_of_size : int Gen.t -> char Gen.t -> string arbitrary
val string : string arbitrary
val string_of : char Gen.t -> string arbitrary
val string_small : string arbitrary
val string_small_of : char Gen.t -> string arbitrary
val string_of_size : int Gen.t -> string arbitrary
val string_printable : string arbitrary
val string_printable_of_size : int Gen.t -> string arbitrary
val string_small_printable : string arbitrary
val string_numeral : string arbitrary
val string_numeral_of_size : int Gen.t -> string arbitrary

(* to-be-deprecated *)
val string_gen : char Gen.t -> string arbitrary
val small_string : string arbitrary
val printable_string : string arbitrary
val printable_string_of_size : int Gen.t -> string arbitrary
val small_printable_string : string arbitrary
val numeral_string : string arbitrary
val numeral_string_of_size : int Gen.t -> string arbitrary
```

`QCheck2.Gen`
---------------------
```ocaml
val bytes_size : ?gen:char t -> int t -> bytes t
val bytes : bytes t
val bytes_of : char t -> bytes t
val bytes_small : bytes t
val bytes_small_of : char t -> bytes t
val bytes_printable : bytes t

val string : string t
val string_of : char t -> string t
val string_small : string t
val string_small_of : char t -> string t
val string_printable : string t

(* to-be-deprecated *)
val small_string : gen:char t -> string t
```

jmid:
@n-osborne can I ask you to take a quick look at the last commits - just one at a time - to see if they make sense - before we merge this PR:
- consistent `QCheck.Gen.{bytes,string}_small{,_of}` combinators - abf86cd
- consistent `QCheck.{bytes,string}_of` combinators - fb1ea0e
- add consistent `QCheck2.Gen.small_{bytes,string}{,_of}` combinators - d1cb85e
- fix comment for `QCheck2.Gen.bytes` - c587c88
- update CHANGELOG - 41ceeae
- add `QCheck.{bytes,string}_small_of` for consistency - 37d8d8a
- update CHANGELOG - a14a767

n-osborne:
Looks good to me. The `*_of` name is indeed nicer than `*_gen` one.

jmid:
Hereby merged - thanks! :smiley:

PR #249, opened 2022-06-29
use Float.equal

n-osborne:
Polymorphic equality is ill-defined regarding `nan`. This PR proposes to use `Float.equal` instead. The rational being that a test failling because `nan <> nan` is not very informative w.r.t the tested program.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 0.19
 
+- use `Float.equal` for comparing `float`s in the `Observable` module underlying function generators.
+
 - add optional `debug_shrink` parameters in alcotest interface and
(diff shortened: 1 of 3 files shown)


jmid:
This looks good to me! :+1: 

I've been trying to come up with a test case to illustrate the fix - and so far failed - so I think it is unfair to ask you to do so... :sweat_smile: 
In any case, it doesn't hurt to compare `nan`s correctly in `Observable`, underlying the tabling of our function generators.

In the CHANGELOG I suggest adding a bit more context to help the reader understand. Here's a proposal:
```diff
- - use `Float.equal` so that `nan` equals `nan`
+ - use `Float.equal` for comparing `float`s in the `Observable` module underlying function generators.
```

PR #253, opened 2022-07-11
Add [no_shrink] to [QCheck2.Gen]

### LINKED ISSUE #232: Disable QCheck2 shrinker with a parameter
That'd be cool to disable the shrinker with a boolean in `Test.make` for instance. I don't know if we can already do this or if it's impossible but this might be a really good feature.

It can happen that the test is wrong and since the generated value is really not-trivial, it takes a lot of time to shrink.

vch9:
Closes #232.

I don't know if that's the most convenient solution, but it seems to work. If you have any suggestion to improve this function, I'd be happy to implement it.


src/core/QCheck2.ml
```diff
@@ -750,6 +750,11 @@ module Gen = struct
   let add_shrink_invariant (p : 'a -> bool) (gen : 'a t) : 'a t =
     fun st -> gen st |> Tree.add_shrink_invariant p
 
+  let without_shrinker (gen: 'a t) : 'a t =
+    make_primitive
+      ~gen:(fun st -> gen st |> Tree.root)
```
vch9, line 755:
```suggestion
      ~gen:generate1
```

I think it's equivalent

jmid: I don't think so, as I suspect `generate1` will ignore cook up its own `Random.State`:
```ocaml
  let generate1 ?(rand=RS.make_self_init()) (gen : 'a t) : 'a =
    gen rand |> Tree.root
```
I think this means the root for a particular run will change if you annotate with `without_shrinker`.
On the other hand, that might be a good test to add... :wink:

vch9: Oh yes, my suggestion was bad indeed. I meant:
```ocaml
make_primitive
      ~gen:(fun rand -> generate1 ~rand gen)
      ...
```

jmid:
Thanks for this! :pray: 

In QCheck(1) there's:
```ocaml
val set_shrink : 'a Shrink.t -> 'a arbitrary -> 'a arbitrary
```
meaning you can achieve something like this with `set_shrink Shrink.nil some_gen`.

Perhaps we should rename and generalize slightly, so that the `QCheck2` operation takes a shrinker function as an argument?
```ocaml
  val set_shrink: ('a -> 'a Seq.t) -> 'a Gen.t -> 'a Gen.t
```
That way
- we continue to keep the interfaces close, in an effort to make QCheck->QCheck2 porting easy
- you should be able to achieve the same functionality with `set_shrink (fun -> Seq.empty) some_gen` (we could also include a binding `Shrink.nil` for the function argument)

jmid:
Looks good!

Three minor/bike-shed'dy things:
- We should probably add a small test of `set_shrink` too?
- Regarding documentation a reader could misunderstand this comment to be side-effecting (if they do not look at the type signature). One possibility is to adapt it to say `returns a generator using [gen] but with shrinking disabled` (or something like that):
  ```ocaml
  val without_shrinker : 'a t -> 'a t
  (** [without_shrinker gen] removes the shrinker from [gen]. *)
  ``` 
- I was thinking about the naming. Looking at names related to shrinking, the last one stands out a bit (only entry using shrink*er*). Furthermore it is a short-hand for something that can be achieved with `set_shrink` - but it is a relatively long identifier. One option could be to use `no_shrink` instead (same form of 'shrink' + short) - I'm open to other proposals though
  ```ocaml
  val add_shrink_invariant : ('a -> bool) -> 'a t -> 'a t
  val set_shrink : ('a -> 'a Seq.t) -> 'a t -> 'a t
  val without_shrinker : 'a t -> 'a t
  ```

jmid:
Sorry, I take back the suggestion of adding a test of `set_shrink`, since you're already using it to implement `without_shrinker`/`no_shrink` and hence testing it as part of that...
(I blame the work injury from having written too many tests of QCheck previously... :sweat_smile:)

vch9:
Hi! Sorry for the huge delay.

Pushed a rebased branch:

1. Used your suggestion for the documentation
2. Went fo `no_shrink`, I like the name, explicit enough and fits well with `set_shrink`
3. Fixed the changelog I think? The `# 0.20` was missing. I don't know if that was intended

jmid:
Looks good to me, thanks!

PR #258, opened 2022-10-07
parametirize string shrinker with a char shrinker

### LINKED ISSUE #257: QCheck.numeral_string may shrink to non-numerals
While reviewing #245 I just realized that some of our `arbitrary string` generators in `QCheck` may have surprising shrinking behavior.
For `QCheck.numeral_string` this means that it may shrink to non-numeral strings:
```ocaml
# let t = Test.make QCheck.numeral_string (fun n -> n <> "" && 5 < int_of_string n);;
val t : QCheck.Test.t = QCheck2.Test.Test <abstr>
# QCheck.Test.check_exn t;;
Exception:
test `anon_test_4` raised exception `Failure("int_of_string")`
on `"a" (after 19 shrink steps)`
# 
```

This is a consequence of how `string_gen` and `string_gen_of_size` falls back on a generic `Shrink.string`
https://github.com/c-cube/qcheck/blob/fa6481fb3ecdc8fe4406ad95c883a136e4cf92f7/src/core/QCheck.ml#L1112-L1128

Generators `numeral_string` and `numeral_string_of_size` are definitely affected.
I suspect that `printable_string` `printable_string_of_size`, and `small_printable_string` may also shrink to strings with non-printable characters under the right conditions.

n-osborne:
This PR is an attempt to fix #257.

It does so by:

- adding an optional argument to `Shrink.string` for character shrinking, with a conservative default
- fix character shrinkers in `QCheck.{printable_string,printable_string_of_size,small_printable_string,numeral_string,numeral_string_of_size}` so that they have the expected behaviour.

For now, we don't expose a `string arbitrary` that takes a shrinker as argument. Albeit it would be nice, there is the question of how to do it. I see three possibilities:

1. add an optional argument to `QCheck.string_gen` and `QCheck.string_gen_of_size`
2. add a non-optional argument to `QCheck.string_gen` and `QCheck.string_gen_of_size`
3. add two functions: `QCheck.string_gen_shrink` and `QCheck.string_gen_shrink_of_size`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changes
 
+- add an optional argument with conservative default to `Shrink.string`
+- fix shrinkers in `QCheck.{printable_string,printable_string_of_size,small_printable_string,numeral_string,numeral_string_of_size}` [#257](https://github.com/c-cube/qcheck/issues/257)
+
 ## 0.19.1
 
 - fix: allow `~count` in `Test.make` to be 0
--- a/src/core/QCheck.ml
+++ b/src/core/QCheck.ml
@@ -770,9 +770,9 @@ module Shrink = struct
     | None -> ()
     | Some shrink -> list_elems shrink l yield
 
-  let string s yield =
+  let string ?(shrink = char) s yield =
     let buf = Buffer.create 42 in
-    list ~shrink:char
+    list ~shrink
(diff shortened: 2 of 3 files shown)


jmid:
Short and sweet - LGTM - thanks! :pray:

PR #259, opened 2022-10-09
Add QCHECK_MSG_INTERVAL environment variable

jmid:
This PR adds an environment variable `QCHECK_MSG_INTERVAL` to control `QCheck_base_runner.time_between_msg`.

This feature is useful in a CI context, where updates are printed on consecutive lines and we want to avoid overflowing the CI log files with too many output lines.

### DIFF
--- a/src/runner/QCheck_base_runner.ml
+++ b/src/runner/QCheck_base_runner.ml
@@ -60,7 +60,14 @@ let set_seed_ ~colors s =
 (* time of last printed message. Useful for rate limiting in verbose mode *)
 let last_msg = ref 0.
 
-let time_between_msg = ref 0.1
+let time_between_msg =
+  let env_var = "QCHECK_MSG_INTERVAL" in
+  let default_interval = 0.1 in
+  let interval = match Sys.getenv_opt env_var with
+    | None -> default_interval
+    | Some f -> float_of_string f in
+  if interval < 0. then invalid_arg (env_var ^ " must be >= 0 but value is " ^ string_of_float interval);
+  ref interval
 
 let get_time_between_msg () = !time_between_msg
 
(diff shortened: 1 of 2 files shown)


jmid:
Rebased on latest `master`.
I've received no objections (or review) to this simple addition. I therefore plan to merge it shortly.

Note: I tested it locally with a simple test with a high count:
```ocaml
open QCheck

let t =
  Test.make ~name:"Interval test" ~count:10_000_000 int (fun i -> i * 2 = i + i)

;;
QCheck_base_runner.run_tests ~verbose:true [t]
```

Running this with variations of `QCHECK_MSG_INTERVAL=0.5 dune exec test/core/intervaltest.exe -- -v` I observed a noticable difference in the update interval compared to, e.g.,  `QCHECK_MSG_INTERVAL=0.1`.

src/runner/QCheck_base_runner.ml
```diff
@@ -60,7 +60,14 @@ let set_seed_ ~colors s =
 (* time of last printed message. Useful for rate limiting in verbose mode *)
 let last_msg = ref 0.
 
-let time_between_msg = ref 0.1
+let time_between_msg =
+  let env_var = "QCHECK_MSG_INTERVAL" in
+  let default_interval = 0.1 in
+  let interval = match Sys.getenv_opt env_var with
+    | None -> default_interval
+    | Some f -> float_of_string f in
```
shym, line 68: What about replacing `float_of_string f` with:
```
match float_of_string_opt f with
| None -> invalid_arg (env_var ^ " must be a float")
| Some i -> i
```
?

jmid: Fair enough, that should give a more descriptive error message to the user. Fixed in efd8f3e

PR #263, opened 2022-11-03
Run CI workflow on OCaml5 too

jmid:


jmid:
CI is all green - so I've squashed the last two commits and will merge afterwards

PR #264, opened 2022-11-03
CI dune runtest on latest MacOSX too

jmid:


jmid:
Merging since CI is happy

PR #271, opened 2023-02-11
Make `Test.check_exn` honor test polarity

alopezz:
Motivation for this PR:

https://github.com/c-cube/qcheck/pull/244#issuecomment-1423824479:

> I was just trying out the library and after seeing [this](https://c-cube.github.io/qcheck/0.20/qcheck-core/QCheck/index.html#tests):
>
> - [Test.make](https://c-cube.github.io/qcheck/0.20/qcheck-core/QCheck/Test/index.html#val-make) to build a test,
>  - [Test.make_neg](https://c-cube.github.io/qcheck/0.20/qcheck-core/QCheck/Test/index.html#val-make_neg) to build a negative test that is expected not to satisfy the tested property,
>   - [Test.check_exn](https://c-cube.github.io/qcheck/0.20/qcheck-core/QCheck/Test/index.html#val-check_exn) to run a single test with a simple runner.
>
> I started to try with some example of my own and I saw that `Test.check_exn` was doing the same for `make` and `make_neg` when using the same property.
>
> E.g. I would have expected the following to raise an exception but it doesn't, which is confusing.
>
> ```ocaml
> let test = Test.make_neg ~name:"int double" small_int (fun i -> i + i = i * 2) in Test.check_exn test
> ```

This PR makes `Test.check_exn` honor the test polarity, making the above example behave in (what I at least consider) a more intuitive manner. A positive side effect of this is that the `ounit` and `alcotest` integration would also no longer need to handle test polarity, removing repetition. To me (admittedly without knowing about the implementation in detail), this looks like a better place to enforce the polarity abstraction. Though it might be possible to take it a step further and move the concept down to the `state` type, I'm not sure if that would make sense.

If this looks like a good idea I'll look into adding tests for this behavior of `check_exn` as necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## NEXT RELEASE
 
+- make `Test.check_result`, `Test.check_cell_exn`, and
+  `Test.check_exn` honor test polarity by raising
+  `Test_unexpected_success` when a negative test (expected to have a
+  counter example), unexpectedly succeeds.
 - ...
 
 ## 0.20
--- a/example/alcotest/output.txt.expected.32
+++ b/example/alcotest/output.txt.expected.32
@@ -29,8 +29,7 @@ on `0 (after 31 shrink steps)`
 │ [FAIL]        suite              4   neg test unexpected success.            │
 └──────────────────────────────────────────────────────────────────────────────┘
 negative test 'neg test unexpected success' succeeded unexpectedly
-ASSERT negative test 'neg test unexpected success' succeeded unexpectedly
-FAIL negative test 'neg test unexpected success' succeeded unexpectedly
+[exception] negative test `neg test unexpected success` succeeded unexpectedly
  ──────────────────────────────────────────────────────────────────────────────
 ┌──────────────────────────────────────────────────────────────────────────────┐
 │ [FAIL]        suite              5   neg fail with error.                    │
--- a/example/alcotest/output.txt.expected.64
+++ b/example/alcotest/output.txt.expected.64
@@ -29,8 +29,7 @@ on `0 (after 63 shrink steps)`
 │ [FAIL]        suite              4   neg test unexpected success.            │
 └──────────────────────────────────────────────────────────────────────────────┘
 negative test 'neg test unexpected success' succeeded unexpectedly
-ASSERT negative test 'neg test unexpected success' succeeded unexpectedly
-FAIL negative test 'neg test unexpected success' succeeded unexpectedly
+[exception] negative test `neg test unexpected success` succeeded unexpectedly
  ──────────────────────────────────────────────────────────────────────────────
 ┌──────────────────────────────────────────────────────────────────────────────┐
 │ [FAIL]        suite              5   neg fail with error.                    │
--- a/example/alcotest/output.txt.expected.ocaml5
+++ b/example/alcotest/output.txt.expected.ocaml5
@@ -29,8 +29,7 @@ on `0 (after 62 shrink steps)`
 │ [FAIL]        suite              4   neg test unexpected success.            │
 └──────────────────────────────────────────────────────────────────────────────┘
 negative test 'neg test unexpected success' succeeded unexpectedly
-ASSERT negative test 'neg test unexpected success' succeeded unexpectedly
-FAIL negative test 'neg test unexpected success' succeeded unexpectedly
+[exception] negative test `neg test unexpected success` succeeded unexpectedly
  ──────────────────────────────────────────────────────────────────────────────
 ┌──────────────────────────────────────────────────────────────────────────────┐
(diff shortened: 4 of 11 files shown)


jmid:
Overall, I agree that the current behavior of `Test.check_exn` may indeed be confusing - and the lack of documentation certainly doesn't help! :grimacing: 

As I indicated in https://github.com/c-cube/qcheck/pull/244#issuecomment-1425616746 I initially put in an effort for users to see the expected counterexample when a negative test fails.
We use this feature extensively in https://github.com/ocaml-multicore/multicoretests to make sure we fail for the expected reason and hence avoid ignoring errors that shouldn't be.
There we use `QCheck_base_runner.run_tests` which builds on `QCheck2.Test.check_cell` https://github.com/c-cube/qcheck/blob/main/src/runner/QCheck_base_runner.ml#L394 and hence isn't affected by this proposed change.

I struggled to get `alcotest` and `ounit` to play nicely and similarly print out the expected negative failure in #244, but the result is nevertheless available for users to inspect IIRC.  On `0.20`, e.g., if I run `dune exec example/alcotest/QCheck_alcotest_test.exe` afterwards I can inspect it:
```
$ cat _build/_tests/my\ test/suite.003.output 

negative test `neg test pass (failing as expected)` failed as expected on: 3 (after 1 shrink steps)
```
If I understand the PR's code correctly, an alcotest user will no longer be able to observe the expected failure of a negative test with the proposed change. Do you agree?
(not to say that this is unsolvable, I just find the potential regression unfortunate :shrug:)

alopezz:
Thanks for looking at this. I wasn't aware of that feature. However, I tested it quickly on my PR and I could also see that message on that output file, so it looks like there's no regression.

Looking closer at the code, I suspect that output is coming from:
https://github.com/c-cube/qcheck/blob/f4a428436451e0463d936dbdf4bcf6eadb108a3c/src/alcotest/QCheck_alcotest.ml#L56

which uses the code from the base runner that checks the details of the result again to recover that information:

https://github.com/c-cube/qcheck/blob/f4a428436451e0463d936dbdf4bcf6eadb108a3c/src/runner/QCheck_base_runner.ml#L152-L155

What do you think?

jmid:
OK, thanks for checking. This puzzled me until I ~~realized~~ recalled that `Raw.callback` has its own polarity printing logic (that I wrote myself in #244 :facepalm:). So at least that mystery has been solved!

Note: because this is expressed as changes to `Test.check_result` which is also used by `Test.check_cell_exn` it leaves us in a situation where 2 out of 3 `Test.check_*` functions in the public interface respect test polarity:
```ocaml
  val check_cell :
    ?long:bool -> ?call:'a callback ->
    ?step:'a step -> ?handler:'a handler ->
    ?rand:Random.State.t -> 'a cell -> 'a TestResult.t

  val check_cell_exn :
    ?long:bool -> ?call:'a callback ->
    ?step:'a step -> ?handler:'a handler ->
    ?rand:Random.State.t -> 'a cell -> unit

  val check_exn : ?long:bool -> ?rand:Random.State.t -> t -> unit
```

This is a bit inconsistent but not necessarily catastrophic as long as it is clearly documented.
Incidentally, I noticed `Test.check_result` is public and hence also needs to have its documentation updated to reflect the change.

alopezz:
As I was thinking of how to document the behavior I implemented here (as per your last response) I realized I'm not convinced that my approach here is an improvement, so it's likely I'll abandon the idea altogether. Thanks for the feedback anyhow!

jmid:
Don't let my review comments discourage you! I'm quite happy to see someone else using the negative test functionality - and I think it's a mess that `Test.check_exn` ignores test polarity.

69589cf and d2b6737 contain a draft proposal for documenting the new behaviour. WDYT?
(This is for `QCheck2` - we should probably document the corresponding functions in the `QCheck` interface too - or give a `QCheck2` pointer)

alopezz:
Thanks for the encouragement and for helping out. My concerns with these changes is that the documentation gets "polluted" with details about a feature which doesn't sound like it should have so much weight.

Note also that this was probably the fragment that led me to be surprised by `check_exn`'s behavior to begin with: https://github.com/c-cube/qcheck/blob/d2b673749149ae58b87ec9d6258174724fe73291/src/core/QCheck2.mli#L1662-L1664

I was planning to revisit this at some point, but I was also thinking of simply adding a note to the documentation of `make_neg` indicating that its polarity is used only on the "runner layer".

I'll have another look soon.

alopezz:
I've added links from the `QCheck` to the `QCheck2` interface and a couple of tests with the new behavior. This is ready for review as far as I'm concerned, as long as you think this is an improvement over the old behavior.

jmid:
Rebased on `main` to reenable CI

jmid:
CI is green and I'm happy with this PR :smiley: 
I've added thus added a CHANGELOG entry in 4ac10da

PR #274, opened 2023-05-01
Realign shrinker tests

jmid:
This little PR realigns the `Shrink` tests to reenable the shrinker benchmark from #177, meaning that the tests should be listed in the same order, and agree on their `name`s.
While adjusting the title of the QCheck2 test, I realized it was testing for a different value than the QCheck(1) test (comparing apples to oranges) so I fixed that.

This whole setup is admittedly fragile, but I'm not bothered enough to fix it ATM.


jmid:
Rebased on `main` after merging #275

jmid:
7 successful CI runs across Ubuntu, macOS and Windows before we hit a web-server failure causing the rest to be cancelled. I'll therefore go ahead and merge.

Another vote for #276

PR #275, opened 2023-05-01
Update GitHub actions

jmid:
Opening #274 I realized the CI was triggered, because of our change from `master` to `main`.
This little PR updates the workflow accordingly, and also adjusts it to test 4.14 and the actual 5.0.0 release.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,10 +2,10 @@ name: build
 on:
   push:
     branches:
-      - master
+      - main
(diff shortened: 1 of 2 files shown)


jmid:
Alcotest 1.7.0 and https://github.com/mirage/alcotest/pull/364 changes the output of a test if `CI=true`.
36e8046 thus adjusts our alcotests to keep the pre-1.7.0-compatible expect test output.

PR #279, opened 2023-05-03
Support 32bit ocaml5 expect tests

jmid:
With OCaml5 gaining traction, opam-ci now has 32-bit OCaml5 targets added. 
This is a combination we previously didn't cover in our expect test setup and this PR thus addresses it.
I discovered while trying to roll a 0.21 release.

I took the opportunity to scrap the initial `dune`-rule generation approach from #250.
With four conditional `copy` rules (one for each OCaml{4,5} x {32,64}-bit combination) we can get something which is relatively simple - as testing code should be.

I dedicate the PR to @c-cube who ~~hated~~ didn't see beauty in the initial `(* -*- tuareg -*- *)` approach :wink: :sweat_smile:


jmid:
While squashing commits I discovered that I must have messed up 64bit outputs, as my 3 updates cancelled each other out... :sweat_smile: 

More seriously, when running a test locally with a runners `example/ounit/run_ounit.sh` or `example/alcotest/run_alcotest.sh` I discovered that `dune` didn't realize the dependencies to the executable files, which would cause failures on the first run - but not on subsequent ones (when the `.exe` was available):
```
File "example/alcotest/dune", line 20, characters 7-27:
20 |  (name QCheck_alcotest_test)
            ^^^^^^^^^^^^^^^^^^^^
./run_alcotest.sh: 1: ./QCheck_alcotest_test.exe: not found
File "example/ounit/dune", line 25, characters 7-24:
25 |  (name QCheck_ounit_test)
            ^^^^^^^^^^^^^^^^^
./run_ounit.sh: 1: ./QCheck_ounit_test.exe: not found
```
For these two, in 9b3745f I therefore reverted to the existing expanded 3-rule approach.

PR #285, opened 2023-09-29
Remove the unnecessary dependency on `bytes`

shym:
If I understand correctly, the `bytes` dependency is really only useful to support older versions of OCaml (before 4.02), but QCheck requires at least 4.08 anyway.
So this PR proposes to drop that dependency altogether.

To give this a bit more context, this PR comes from wanting to run some tests that use QCheck core in a CI setup where Opam is not an option (yet). So this is working by vendoring QCheck and relying on dune’s modularity. For some reason, dune resolves correctly the `unix` library required in `src/core/dune` but not `bytes`. With this PR, the dependencies of that CI setup drop to only OCaml and dune.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## NEXT RELEASE
 
-- ...
+- Drop the dependency on `base-bytes` as it is provided in all supported
+  versions of the OCaml compiler
 
 ## 0.21.2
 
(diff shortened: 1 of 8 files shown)


jmid (approved):
LGTM, thanks!
Should we add a CHANGELOG entry to let end-users know of the change? :thinking:

shym:
I can add an entry (I would add a new “NEXT RELEASE” entry for that, I suppose?). But, as it _should_ be really transparent to users (removing dependencies is easy), would that be a notable change?

jmid:
If nothing else, I suppose it could be handy to keep track of which version dropped the dependency :shrug:

shym:
I added an changelog entry and rebased on `main`.

Issue #281, opened 2023-07-13
Documentation - role of `Gen.t` vs. `arbitrary`


lindig:
I am picking up QCheck and I struggle with how to write my tests; I believe the documentation would benefit from describing the relation of generators and arbitrariness better because they have a similar structure. For example: I need a small list of key-value pairs. Should I use the `Gen.t` or `arbitrary` combinators to define this? Some  `arbitrary` functions take `Gen.t` values as arguments - which is surprising and creates the confusion.

jmid:
Acknowledged! :slightly_smiling_face: 

In a past role teaching this stuff I prepared slides to explain the `Gen.t` vs `arbitrary` difference.
See https://janmidtgaard.dk/quickcheck/slides/lec03.pdf from p.48 of the pdf (labeled 21/48 in the bottom right corner).
The slides summarize my own view: `Gen.t` is purely for generation and nothing else, while `arbitrary` is a generator packaged up with a corresponding printer, shrinker, etc.

With that in mind, e.g., a `string arbitrary` combinator needing a 'size generator' we really only need "something that can generate an `int` size", i.e., an `int Gen.t`. 

Whereas my old slides go top-down, the relevant Cornell book chapter instead explains this bottom-up, starting from "pure generation" and explains `arbitrary` as their enhancement IIRC.

Finally I should mention that `QCheck2` with integrated shrinking, doesn't have the same distinction because generators and shrinkers are merged into one common (lazy tree) structure.

> For example: I need a small list of key-value pairs. Should I use the Gen.t or arbitrary combinators to define this?

Tests built with `Test.make` accept an `arbitrary` in order to be able to print (and shrink) any counterexamples.
As such I expect you would want something like the following (using `char` keys and `bool` values):
```ocaml
list_of_size Gen.small_nat (pair char bool) : (char * bool) list QCheck.arbitrary =
```
Notice how this uses `Gen.t` for size generation, for the reason explained above.


I hope this helps! As always, PRs to improve the documentation are very welcome! :pray::smiley:

lindig:
The documentation on the slides is  helpful; maybe include a link or the slides in this repository. I believe this kind of high-level explanation would at the top of the MLI file would help. Something that is not obvious: is `QCheck2` an older or more recent version? The current release is `0.21.x` so I fail to see the connection and which interface should new code target?

jmid:
[...]

PR #286, opened 2024-01-20
Add sig expander to qcheck2

mbarbin:
This PR adds support for the `[@@deriving qcheck2]` construct in signatures (such as in `*.mli` files).

For a motivating example, see [here](https://github.com/mransan/ocaml-protoc/pull/233).

Thank you!

### DIFF
--- a/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
+++ b/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
@@ -634,6 +634,27 @@ let gen_expander_qcheck = Deriving.Generator.V2.make_noarg create_arbs
 
 let gen_expander_qcheck2 = Deriving.Generator.V2.make_noarg (create_gens `QCheck2)
 
-let _ = Deriving.add "qcheck" ~str_type_decl:gen_expander_qcheck
+let gen_sig_expander_qcheck2 =
+  Deriving.Generator.V2.make_noarg
+    (fun ~ctxt  (_rec_flag, tds) ->
+      let loc = Expansion_context.Deriver.derived_item_loc ctxt in
+      let (module A) = Ast_builder.make loc in
+      tds |> List.map (fun td ->
+        let type_name = td.ptype_name.txt in
+        A.psig_value
+          (A.value_description
+            ~name:(A.Located.mk (name type_name))
+            ~type_:(A.ptyp_constr (A.Located.mk (Ldot (Ldot (Lident "QCheck2", "Gen"), "t")))
+                                    [ A.ptyp_constr (A.Located.mk (Lident type_name)) [] ])
+            ~prim:[])))
 
-let _ = Deriving.add "qcheck2" ~str_type_decl:gen_expander_qcheck2
+let _ =
+  Deriving.add
+    "qcheck"
+    ~str_type_decl:gen_expander_qcheck
+
+let _ =
+  Deriving.add
+    "qcheck2"
+    ~str_type_decl:gen_expander_qcheck2
+    ~sig_type_decl:gen_sig_expander_qcheck2
--- a/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
+++ b/src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
@@ -636,7 +636,7 @@ let gen_expander_qcheck2 = Deriving.Generator.V2.make_noarg (create_gens `QCheck
 
 let gen_sig_expander_qcheck2 =
   Deriving.Generator.V2.make_noarg
-    (fun ~ctxt  (_rec_flag, tds) ->
+    (fun ~ctxt (_rec_flag, tds) ->
       let loc = Expansion_context.Deriver.derived_item_loc ctxt in
       let (module A) = Ast_builder.make loc in
       tds |> List.map (fun td ->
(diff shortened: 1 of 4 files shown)


test/ppx_deriving_qcheck/deriver/qcheck2/test_sig.ml
```diff
@@ -0,0 +1,42 @@
...
+let test_string () =
+  test_compare ~msg:"Gen.string <=> deriving string" ~eq:Alcotest.string Gen.string T.gen_string'
+
+(** {2. Execute tests} *)
+
+let () = Alcotest.run "Test_Primitives"
+           [("Primitives",
+             Alcotest.[
+                 test_case "test_int" `Quick test_int;
+                 test_case "test_unit" `Quick test_string;
```
jmid, line 41: All 6 test titles here seem copy-pasted from `test_primitives.ml` (which is a natural starting point).
To separate their outcome in the log from the other ones, it would make sense to include the keyword `"signature"` in them.

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
```diff
@@ -634,6 +634,27 @@ let gen_expander_qcheck = Deriving.Generator.V2.make_noarg create_arbs
 
 let gen_expander_qcheck2 = Deriving.Generator.V2.make_noarg (create_gens `QCheck2)
 
-let _ = Deriving.add "qcheck" ~str_type_decl:gen_expander_qcheck
-
-let _ = Deriving.add "qcheck2" ~str_type_decl:gen_expander_qcheck2
+let gen_sig_expander_qcheck2 =
+  Deriving.Generator.V2.make_noarg
+    (fun ~ctxt  (_rec_flag, tds) ->
```
jmid, line 639: nit: double space

jmid (changes requested):
Thanks for your PR - and sorry for the delay. 

I was a little puzzled about the tests, only checking `type _ = ...` cases in the signature since that breaks abstraction, until I realized that this may not be the fault of the PR:  a type equation such as `type t = int` is actually promoted to be exposed in the signature by ppx_deriving, when using an 'inline signature' like the PR's test does, despite starting with an abstract `type t` input... :thinking: 

For this reason and also because I think it helps readabilty of the derived signatures, I suggest splitting the test module signature and implementation:
```ocaml
module type TSig = sig
  type t [@@deriving_inline qcheck2][@@@deriving.end]
  type string' [@@deriving_inline qcheck2][@@@deriving.end]
  type 'a tt [@@deriving_inline qcheck2][@@@deriving.end]
end

module T : TSig = struct
  type t = int [@@deriving qcheck2]
  type string' = string [@@deriving qcheck2]
  type 'a tt = 'a list [@@deriving qcheck2]
end
```

I think it would be nice to confirm with tests, that the derived signatures work for all primitive types (unit, char, int, int32, ...). For example, `Gen.option` takes an optional `ratio` parameter thus changing its signature. Is it obvious that this works out?

The PR unfortunately breaks down in the presence of type parameters, such as in the `'a tt` case above. Here I think we want to derive a function accepting a generator for each type parameter: `'a Gen.t -> 'a tt Gen.t` (suitably generalized).

I haven't used `@@deriving_inline` much myself before. From https://ocaml.org/docs/metaprogramming#dropping-ppxs-dependency-with-deriving_inline I understand that the printed outcome may differ between compiler versions. Do you have a sense of how stable the test output is across OCaml 4.08.0-5.2.0? Is `runtest` sufficient to keep the derived generator test code up to date or should we go for the `@lint` target mentioned in the above?

Finally, for symmetry I would have liked to see `QCheck`(1) support too, but I think this is too much to require. If you can get `QCheck2` to work, I may take a stab at that afterwards.

mbarbin:
Thanks @jmid for your review!

Looking back, I'm not sure why I left the `deriving_inline` construct here. I used it to track generated code while working on the PR, but I've since learned about `dune show pp path/to/file.ml`. This construct isn't tied to this PR, so I've removed it.

You pointed out parametrized types and Gen.option, which I hadn't thought of. I need to study these more, so I'm changing the PR to a draft. Thanks again!

src/ppx_deriving_qcheck/ppx_deriving_qcheck.ml
mbarbin: Thanks, fixed!

test/ppx_deriving_qcheck/deriver/qcheck2/test_sig.ml
mbarbin: Ah, I am not familiar with alcotest, and had missed the naming pattern in use. I've made an attempt to improve based on your comment. Thank you!

jmid:
In #281 there's a reasonable explanation of the QCheck / QCheck2 relation, incl. additional pointers.

mbarbin:
Hello,

I wanted to provide an update regarding supporting the [@@deriving qcheck2] construct in signatures.

After further consideration and exploration of my project requirements, I've decided to take a different approach that doesn't necessitate this feature. As such, I find it difficult to justify the time needed to address the issues you've pointed out and complete the implementation.

I regret not being able to finish the work I started and I apologize for any inconvenience this may cause. I want to express my gratitude for the valuable feedback and insights you've provided during our discussions. I am hopeful this information will be beneficial if anyone picks up this project in the future.

I'm now inclined to close this PR. Perhaps we could turn it into an open issue instead? I hope this decision is understandable and I look forward to potential collaborations in the future.

Thank you for your time and understanding.

Issue #287, opened 2024-06-21
memory usage problems


Robotechnic:
Hello,
I started using qcheck for a project and it's a great quick checking framework but I have a problem with the memory consumption on some tests which require to manipulate huge data structures or a lot of data:
```ml
open QCheck2

let generator = Gen.(
  (0 -- 500) >>= fun n ->
  tup2 (return n) (list_size (return n) (array_size (1 -- 200) int))
)

let test = Test.make ~count:1000 ~name:"test" generator (fun _ ->
  true)

let () = exit @@ QCheck_runner.run_tests [test;test;test]
```
This is a minimal example and my problem is that it uses 1.1gb of RAM if I run it like that:
```ml
let () = ignore @@ QCheck_runner.run_tests [test]
let () = ignore @@ QCheck_runner.run_tests [test]
let () = exit @@ QCheck_runner.run_tests [test]
```
It takes the same amount of time to run but uses 438Mb of RAM, which is what I would expect from a test.
Here the example is dumb but in my case, I'm running 500 tests that run fine because they don't use much RAM but I have 30 tests that uses a lot of RAM and running them all at once consumes more than 10gb of RAM.

Robotechnic:
I forgot to gave some precisions about my config:
It's done using ocaml 4.14.2 and check 0.21.3

c-cube:
A quick calculation shows that in the worst case, each array could be
500 * 200 * 8B = 800kB of data. On average it might be 250 * 100 * 8B
= 200kB. If you run 1000 tests, that means you might consume 200kB *
1000 = 200MB, plus some other overhead from qcheck itself. That's not
insignificant, although I'm not sure how it gets to 1.1GB even being
allocated (perhaps a lot of closures and temporary values used to
produce the values?).

I don't think it's super shocking, 500*200 is fairly big for nested
arrays :)

Maybe run `Gc.full_major()` at the beginning of some tests? Or with a
low random probability?

ghilesZ:
Just passing by and interested by this. Here are two ideas.

- Maybe try with QCheck (instead of QCheck2)
- Maybe you can try writing a custom array generator that recycles the arrays the next time an array of the same size is drawn.  In this case you would save the allocation time and only have to pay for the initialization. Of course, this is only useful when you generate enough arrays that collisions are likely.
  
It would look roughly like that : (I'm unsure how to do the same with QCheck2 since the `Gen.t` type is not public anymore)

````OCaml
open QCheck

let array_memo_size =
[...]

Issue #51, opened 2018-03-30
Access instances of a test case


c-cube:
even if the test succeeded, we should store and give access to the instances (eg. for printing, if the user wishes so)

PR #288, opened 2024-06-28
Remove QCheck2.TestResult.get_instances to address memory leak

jmid:
`QCheck.TestResult.t.instances` was originally added in dc53caf to address #51.
With the addition of `QCheck2` it was moved there, `QCheck.TestResult.t` was made abstract, and instead
```ocaml
  val QCheck2.TestResult.get_instances : 'a QCheck2.TestResult.t -> 'a list
```
was exposed in the 0.18 interface.

Now https://sherlocode.com/?q=get_instances reveals that noone uses the binding, however #287 reveals that keeping all test inputs causes memory leaks
- that increase with the test count
- that accumulate further as a test list grows.

This PR therefore proposes to remove the list of instances from `TestResult`.

With the PR, on the test case from @esope QCheck2 (and QCheck) regains constant memory usage across increasing test counts - a nice property IMO (beware slight variation below due to different seeds):
```
$ /usr/bin/time --format="max memory: %M kB" _build/default/test_esope.exe 10000
random seed: 404829644
success (ran 1 tests)
max memory: 12628 kB
$ /usr/bin/time --format="max memory: %M kB" _build/default/test_esope.exe 20000
random seed: 234774561
success (ran 1 tests)
max memory: 12500 kB
$ /usr/bin/time --format="max memory: %M kB" _build/default/test_esope.exe 30000
random seed: 263321469
success (ran 1 tests)
max memory: 12628 kB
```
In addition, 12.6MB also seems more reasonable compared to the 300MB, 583MB, 881MB the above use to take!
I suspect performance will improve a bit, due to the decreased GC stress.

For the 2 cases reported by @Robotechnic the situation improves nicely too:
```
$ /usr/bin/time --format="max memory: %M kB" _build/default/robotechnic_list.exe
random seed: 243559907
success (ran 3 tests)
max memory: 64120 kB
$ /usr/bin/time --format="max memory: %M kB" _build/default/robotechnic_sep.exe
random seed: 490585175
success (ran 1 tests)
success (ran 1 tests)
success (ran 1 tests)
max memory: 64928 kB
```

To recap, in a list of 3 tests this uses ~64MB down from ~1151MB
whereas 3 separate runs uses ~65 MB down from ~438MB (again ignoring different seeds)

Because of the way we process tests:
- run and print pass/fail as they execute
- secondly print counterexamples and statistics afterwards

we cannot realistically obtain constant memory usage as test lists grow.

However, as the above numbers show, this seems much more acceptable in the light of the overall memory reduction this PR offers.

----

Edit - Addition:
I believe this will make an even bigger difference for QCheck2 test than for QCheck ones, as the former's generator produces larger lazy shrink trees.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## NEXT RELEASE
 
-- ...
+- Remove `QCheck2.TestResult.get_instances` as retaining previous test inputs
+  cause memory leaks
 
 ## 0.21.3
 
--- a/src/core/QCheck2.ml
+++ b/src/core/QCheck2.ml
@@ -1354,9 +1354,6 @@ module TestResult = struct
     collect_tbl: (string, int) Hashtbl.t lazy_t;
     stats_tbl: ('a stat * (int, int) Hashtbl.t) list;
     mutable warnings: string list;
-    mutable instances: 'a list;
-    (** List of instances used for this test, in no particular order.
-        @since 0.9 *)
   }
 
   let get_state {state; _} = state
@@ -1390,8 +1387,6 @@ module TestResult = struct
   let get_warnings r = r.warnings
 
   let warnings = get_warnings
-
-  let get_instances r = r.instances
 
   let is_success r = match r.state with
     | Success -> true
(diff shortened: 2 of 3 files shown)


jmid:
Note: the CI failure is unrelated and should be addressed by  #289

jmid:
Rebased on `main` after merging #289

PR #291, opened 2024-07-01
Make QCheck2.state.res immutable

jmid:
QCheck2 fails with compile time warnings:
```ocaml
File "src/core/QCheck2.ml", line 1570, characters 4-33:
1570 |     mutable res: 'a TestResult.t;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning 69 [unused-field]: mutable record field res is never mutated.
```

I think this was just an oversight with the refactoring in connection with QCheck2.
The PR removes the warning and the test suite still passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Remove `QCheck2.TestResult.get_instances` as retaining previous test inputs
   cause memory leaks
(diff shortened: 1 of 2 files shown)


jmid:
Rebased on `main` and added a CHANGELOG entry

PR #292, opened 2024-07-04
Update CI to setup-ocaml v3

jmid:
... and add a test on 5.2


jmid:
Removed attempt to enable test suite run under Windows, as it is was hitting an expect test difference - reported in #293

PR #295, opened 2024-11-21
PPX: mutual recursion with records is now compiling.

Kakadu:
Previously, the code like below was not compiling because 'n' argument was being forgotten. I added a dirty hack of treating all types in mutual block as recursive. Don't know is it a right solution but it seems to help

```ocaml
type tree = Leaf | Node of  tree * tree
and name  = { a: tree }
[@@deriving qcheck2]
```


jmid:
LGTM! FTR, I consider this a natural fix for the issue - and it even comes as a slight code simplification.

Merging ... thanks!

PR #296, opened 2024-12-06
Quote and escape Observable strings and chars

jmid:
When recently using QCheck's function generation support, I noticed an issue with its printing of chars and strings.
Consider these two examples:
```ocaml
open QCheck

let test_char =
  Test.make
    (triple char char (fun1 Observable.char bool))
    (fun (c1,c2,f) -> Fn.apply f c1 = Fn.apply f c2)

let _ = QCheck_base_runner.run_tests ~verbose:true [test_char]

let test_string =
  Test.make
    (triple string string (fun1 Observable.string bool))
    (fun (s1,s2,f) -> Fn.apply f s1 = Fn.apply f s2)

let _ = QCheck_base_runner.run_tests ~verbose:true [test_string]
```

resulting in output such as:
```
generated error fail pass / total     time test name
[✗]    1    0    1    0 /  100     0.0s anon_test_1

--- Failure --------------------------------------------------------------------

Test anon_test_1 failed (15 shrink steps):

('a', '\142', {� -> true; _ -> false})
================================================================================
failure (1 tests failed, 0 tests errored, ran 1 tests)
generated error fail pass / total     time test name
[✗]    2    0    1    1 /  100     0.0s anon_test_2

--- Failure --------------------------------------------------------------------

Test anon_test_2 failed (27 shrink steps):

("a", "", { -> false; _ -> true})
```

Note the not-so-helpful function domain printing: `� -> true` and ` -> false`.

This PR therefore offers a quickfix to these two printers, aligning them with the `arbitrary` printers, and thus offering a uniform print out:
```
generated error fail pass / total     time test name
[✗]    1    0    1    0 /  100     0.0s anon_test_1

--- Failure --------------------------------------------------------------------

Test anon_test_1 failed (18 shrink steps):

('a', '_', {'_' -> false; _ -> true})
================================================================================
failure (1 tests failed, 0 tests errored, ran 1 tests)
generated error fail pass / total     time test name
[✗]    1    0    1    0 /  100     0.0s anon_test_2

--- Failure --------------------------------------------------------------------

Test anon_test_2 failed (34 shrink steps):

("a", "", {"" -> false; _ -> true})
================================================================================
failure (1 tests failed, 0 tests errored, ran 1 tests)
```

Looking further,
- I was annoyed by having to duplicate the ad-hoc printers from `arbitrary` one more time, rather than use the `Print` module, supposed to be used for this
[...]

PR #297, opened 2024-12-06
Quote and escape in QCheck.Print.{char,string}, aligning it with QCheck2.Print

jmid:
As mentioned in #296 
- the `QCheck.Observable.{char,string}` function printers yield unhelpful output, despite using `QCheck.Print.{char,string}`
- quickfixing it like #296, adds another ad-hoc copy, rather than fixing the root cause, namely `QCheck.Observable.{char,string}`
- `QCheck2.Print.{char,string}` already quotes and escapes, creating a `QCheck.Print` / `QCheck2.Print` mismatch for chars and strings

This PR thus suggests the breaking change of fixing `QCheck.Print.{char,string}`.
This further opens up for cleaning up some of the ad-hoc printers in the `arbitrary` combinators (commit d1c1fee)

I'm hoping for fellow QCheck users and developers to chip in to the discussion and share which PR and `QCheck2.Print.{char,string}` they prefer... :pray: :slightly_smiling_face:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## NEXT RELEASE
 
+- Quote and escape in `Print.string` and `Print.char` in the `QCheck` module,
+  mirroring the `QCheck2.Print` module's behaviour
 - Clean-up `QCheck` and `QCheck2` documentation pages
 - Add `exponential` generator to `QCheck`, `QCheck.Gen`, and `QCheck2.Gen`
 - Add `Shrink.bool` and use it in `QCheck.bool`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## NEXT RELEASE
 
 - Quote and escape in `Print.string` and `Print.char` in the `QCheck` module,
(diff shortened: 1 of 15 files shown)


jmid:
To assess if this change would break anything I've rolled an opam-repo PR ocaml/opam-repository#27038
The conclusion is clear: there were (the usual) failures but none caused by `QCheck.Print.{char,string}` printers changing to properly quote and escape. 

Based on this experiment I therefore favor this one over #296. OK with you @c-cube?

jmid:
Rebased on `main` after merging #298

jmid:
Looking again at (our workaround in multicoretests)[https://github.com/ocaml-multicore/multicoretests/blob/main/lib/STM.ml#L28-L34] I was reminded that `bytes` has the same problem - in both `QCheck.Print` and in `QCheck2.Print`.

The latest commits thus fix `Print.bytes` too.

The expect test updates in d8f3b5f and f773f75 document why this helps the end-user output.

jmid:
Rebased on main after merging #302

jmid:
OK. Since no objections have surfaces I'll go ahead and merge this one.

PR #298, opened 2024-12-06
Add an exponential generator

jmid:
This PR adds an exponential generator to both QCheck and QCheck2, using a mapping from a uniform [0;1] distribution
https://en.wikipedia.org/wiki/Relationships_among_probability_distributions
https://en.wikipedia.org/wiki/Exponential_distribution#Random_variate_generation

All of them takes a mean/average parameter, e.g., `exponential 10.`, and results in a distribution where samples approximate that mean/average:
```
+++ Stats for exponential 10. dist ++++++++++++++++++++++++++++++++++++++++++++++++++++++++

stats dist:
  num: 5000, avg: 9.56, stddev: 9.91, median 7, min 0, max 89
    0..  4: #######################################################        1934
    5..  9: ##################################                             1202
   10.. 14: ####################                                            727
   15.. 19: ############                                                    452
   20.. 24: ########                                                        284
   25.. 29: ####                                                            164
   30.. 34: ##                                                              103
   35.. 39: #                                                                51
   40.. 44:                                                                  26
   45.. 49:                                                                  24
   50.. 54:                                                                  15
   55.. 59:                                                                   7
   60.. 64:                                                                   3
   65.. 69:                                                                   2
   70.. 74:                                                                   2
   75.. 79:                                                                   1
   80.. 84:                                                                   1
   85.. 89:                                                                   2
   90.. 94:                                                                   0
   95.. 99:                                                                   0
```
We've found the combinator useful in QCheck-STM to generate `cmd` lists of a reasonable length ocaml-multicore/multicoretests#472 but thought the combinator could be more generally useful.

We may even consider switching our `list`, `array`, `string`, and `bytes` generators to output lengths following such an exponential distribution, but that will be for another day - and PR...


jmid:
Rebased on main after merging #299

PR #299, opened 2024-12-06
Add Shrink.bool

### LINKED ISSUE #251: Missing QCheck.Shrink.bool
I noticed that QCheck.Shrink is missing a `bool` combinator.

This should be probably shrink `true` to `false` like QCheck2: 
  https://github.com/c-cube/qcheck/blob/dd5feb49be06b585b943037e7a6c218d9cb7fb49/src/core/QCheck2.ml#L313-L317

Finally the `bool arbitrary` could be extended to use it.

jmid:
This PR adds the missing Shrink.bool combinator and thus fixes #251

Like the QCheck2 bool shrinker it shrinks towards false (for uniformity)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## NEXT RELEASE
 
+- Add `Shrink.bool` and use it in `QCheck.bool`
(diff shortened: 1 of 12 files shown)


jmid:
Rebased on `main` after merging #300

PR #307, opened 2024-12-27
Add missing int32 and int64 combinators

jmid:
This PR adds a bunch of missing `int32` and `int64` combinators:
- `Print.int{32,64}` to both `QCheck` and `QCheck2`,
- `QCheck.Gen.int{32,64}`, and
- `Observable.int{32,64}` to both `QCheck` and `QCheck2`

Simultaneously the PR deprecates `QCheck.Gen.{ui32,ui64}` like is already done for `QCheck2.Gen.{ui32,ui64}` for uniformity.

Fixes #304
Fixes #301
Fixes #90

PR #316, opened 2025-01-21
Add missing tests of int{32,64}, option, and result combinators

### LINKED ISSUE #310: Add tests for new combinators
#307 and #308 adds new combinators.

We should add tests of them
- unit tests for shrinkers
- negative property tests (of false properties) to exercise generators
- stat distribution tests of generators
- ...

jmid:
Fixes #310 

#307 and #308 added new combinators. This PR adds tests of them:

- unit tests for shrinkers
- positive property tests
- negative property tests (of false properties) to exercise generators
- stat distribution tests of generators

and even a couple of function tests of `int{32,64}`,


jmid:
All green and proof-read. Merging.

Issue #163, opened 2021-08-18
Sub-optimal QCheck2 function shrinkers


jmid:
**Sub-optimal function shrinkers**

A number of the function tests illustrate a difference in the function shrinker's output (terminal cuts off at width 160):
```
--- Failure -----------------------------------------------------------------	--- Failure -----------------------------------------------------------------

Test fail_pred_map_commute failed (127 shrink steps):			      |	Test fail_pred_map_commute failed (16 shrink steps):

([3], {_ -> 0}, {3 -> false; _ -> true})				      |	([2], {_ -> 0}, {1 -> false; 2 -> true; _ -> false})

...

--- Failure -----------------------------------------------------------------	--- Failure -----------------------------------------------------------------

Test fold_left fold_right failed (25 shrink steps):			      |	Test fold_left fold_right failed (22 shrink steps):

(0, [1], {(1, 0) -> 1; _ -> 0})						      |	(0, [1], {(1, 0) -> 1; (8, 0) -> 0; (8, 8) -> 0; (8, 93) -> 0; (7, 7) -> 0; (

...

--- Failure -----------------------------------------------------------------	--- Failure -----------------------------------------------------------------

Test fold_left fold_right uncurried failed (111 shrink steps):		      |	Test fold_left fold_right uncurried failed (325 shrink steps):

({(5, 7) -> 0; _ -> 7}, 0, [5; 0])					      |	({(23, 62) -> 0; (9, 42) -> 0; (8, 61) -> 0; (8, 5) -> 0; (30, 5) -> 0; (9, 6

--- Failure -----------------------------------------------------------------	--- Failure -----------------------------------------------------------------

Test fold_left fold_right uncurried fun last failed (26 shrink steps):	      |	Test fold_left fold_right uncurried fun last failed (25 shrink steps):

(0, [1], {(0, 1) -> 1; _ -> 0})						      |	(0, [1], {(0, 2) -> 0; (8, 80) -> 0; (93, 9) -> 0; (7, 24) -> 0; (8, 0) -> 0;

```
The steps spent vary between the two approaches. The last two are perhaps most saying.
First off, here's the counterexamples reported by QCheck2:
```
--- Failure --------------------------------------------------------------------

Test fold_left fold_right uncurried failed (325 shrink steps):

[...]

Issue #157, opened 2021-08-17
Improve QCheck2 string shrinking


sir4ur0n:
**Sub-optimal string shrinkers**

By design, with integrated shrinking a shrinker inherits the generator's structure. As a consequence, the reduced counterexamples are not as small as previously. Here's `diff -y` of `string_never_has_000_char` and `string_never_has_255_char`:
```
--- Failure -------------------------------------------------	--- Failure -------------------------------------------------

Test string never has a \000 char failed (25 shrink steps):   |	Test string never has a \000 char failed (22 shrink steps):

"\000"							      |	"aaaaaa\000aaaaaaaaaaaaaaaa"

--- Failure -------------------------------------------------	--- Failure -------------------------------------------------

Test string never has a \255 char failed (249 shrink steps):  |	Test string never has a \255 char failed (59 shrink steps):

"\255"							      |	"aaaaaaaaaaaaaaaaaaaaaaaaaa\255aaaaaaaaaaaaaaaaaaaaaaaa"
```

Since `Gen.string` ends up calling `Gen.bytes_size` - and since that generates a random `size` first and subsequently a string of that desired `size`, I believe the shrinker first investigates the shrink tree of `size` to find the smallest one to fail the property, and then fixes that while reducing the individual characters. This can be confirmed with a shrink log (here I just list successful string shrinks):
```
"\011\241\027<r\r\130\187\151\200'\000\181,\019qi7\193\029\195\218L\210\186_"
"V\020\214t\235a\000hl\014z\200\243,\227\171Sg\172\211\244\183F"
"a\020\214t\235a\000hl\014z\200\243,\227\171Sg\172\211\244\183F"
"aa\214t\235a\000hl\014z\200\243,\227\171Sg\172\211\244\183F"
"aaat\235a\000hl\014z\200\243,\227\171Sg\172\211\244\183F"
"aaaa\235a\000hl\014z\200\243,\227\171Sg\172\211\244\183F"
"aaaaaa\000hl\014z\200\243,\227\171Sg\172\211\244\183F"
"aaaaaa\000al\014z\200\243,\227\171Sg\172\211\244\183F"
"aaaaaa\000aa\014z\200\243,\227\171Sg\172\211\244\183F"
... [some line omitted]
"aaaaaa\000aaaaaaaaaaaaa\244\183F"
"aaaaaa\000aaaaaaaaaaaaaa\183F"
"aaaaaa\000aaaaaaaaaaaaaaaF"
[...]

Issue #268, opened 2022-11-30
Non-minimal list counter examples


jmid:
In some (relatively rare) circumstances the `QCheck.Shrink.list_spine` shrinker can produce non-minimal counterexamples.
Notice the last 4-element counterexample in the following:
```ocaml
# #require "qcheck-core";;
# #require "qcheck-core.runner";;
# open QCheck;;
# let t =
  Test.make
    (small_list small_nat)
    (fun xs -> not (List.mem 0 xs && List.mem 1 xs && List.mem 2 xs));;

# QCheck.Test.check_exn ~rand:(Random.State.make [|16;17;16|]) t;;
Exception:
test `anon_test_1` failed on ≥ 1 cases: [1; 0; 2] (after 13 shrink steps)
                                          
# QCheck.Test.check_exn ~rand:(Random.State.make [|16;17;17|]) t;;
Exception:
test `anon_test_1` failed on ≥ 1 cases: [1; 0; 2] (after 2 shrink steps)
                                          
# QCheck.Test.check_exn ~rand:(Random.State.make [|16;17;18|]) t;;
Exception:
test `anon_test_1` failed on ≥ 1 cases: [2; 1; 0; 0] (after 4 shrink steps)
```

This is a consequence of the computationally fast list shrinker from #242.
In slogan form the issue can be phrased as: "speed or exhaustiveness - pick one".

In a few more words:
- A naive shrinker will try to remove each element in turn, producing O(n) shrinking candidates
  - Even that strategy is not exhaustive, as some reductions requires simultaneously removing more than one element (occurrences of `add k; remove k` that cancel out each other is one such example)
  - It will be prohibitively slow on longer lists though
  - We previously had a slower, but more exhaustive shrinker (see #64 for more details)
- The current list shrinker is faster and thus scales to larger lists
  - To do so, we should only produce O(log n) shrinker candidates on a single shrinker invocation
  - This prohibits trying all removal candidates
  - The current shrinker is motivated and explained in more detail in #235 

I'm not sure what would be the best fix to the issue without breaking the computational complexity.

In https://github.com/jmid/multicoretests where we hit this, I'm experimenting with simply stopping the recursion a bit earlier and emitting hand-chosen shrinker candidates for the new base cases.
```ocaml
    let rec shrink_list_spine l yield =
      let rec split l len acc = match len,l with
        | _,[]
        | 0,_ -> List.rev acc, l
        | _,x::xs -> split xs (len-1) (x::acc) in
      match l with
[...]

PR #319, opened 2025-01-31
QCheck2 shrinker improvements to list, array, bytes, strings, and functions

jmid:
This PR takes a first step towards improving the QCheck2 list shrinker.
The current one uses the input size tree and offers only prefixes, e.g., shrinking [1;2;3;4;5;6;7;8] to [], [1;2;3;4], [1;2;3;4;5;6], [1;2;3;4;5;6;7] (or something like that). This results in needlessly large counterexamples, if the property is essentially "lists do not contain the int 7". :grimacing: 

The PR replaces the list shrinker with a recursive one instead shrinking [1;2;3;4;5;6;7;8] to
[1;2;3;4], [5;6;7;8], [2;3;4;5;6;7;8] or [1;2;3;4;6;7;8], i.e., 
- trying to keep either half or
- dropping the head element from either half

The shrinker then continues size reduction recursively from either of these, and switches to element shrinking afterwards.
The approach is not perfect, but it is a good step forward over the current status, and together with #318 it gets QCheck2 working  better (witness the differences in expect test outputs).

Secondly the PR switches the other list, array, bytes, string #157, and function generators/shrinkers #163 to use the improved list shrinker, so that these benefit from the improvement too.

The PR does not improve the `QCheck2.Gen.{list_size, array_size, bytes_size, string_size}` generators. 
These take a size generator parameter, and I've not figured out how to cleanly respect that (read: preserve invariants), e.g., if a user provides a size generator of "even numbers", "multiples of 7", or "prime numbers". As such,
the shrinking behaviour and performance of those remain untouched.

QCheck2 still fares badly on < ocaml.5.0 because it needs `Random.State.split` and on OCaml4 that builds on a hack - that runs pretty slowly (on OCaml 5+ it works nice and fast). I have another patch to replace the hack with a faster one, that I'll submit as a separate PR.

The PR adds a bunch of QCheck2 shrinker unit tests to document the effect of the change.
Should anyone wish to, the PR can thus be read commit-by-commit as gradual improvements.

I'm keen to hear from any QCheck2 users if this helps them (polite ping to @vch9 and friends)


jmid:
> The approach is not perfect, but it is a good step forward over the current status, and together with #318 it gets QCheck2 working better (witness the differences in expect test outputs).

FTR, the proposed heuristic fails to reduce a test input [1;2;3;4;5;6;7;8] if the tested property is something like
"int lists do not contain both 1 and 5", as none of [1;2;3;4], [5;6;7;8], [2;3;4;5;6;7;8] or [1;2;3;4;6;7;8] are counterexamples to it.

This is a more rarely occurring limitation over "shrink to prefixes" (witness again the differences in expect test outputs), as it requires a combination of
- two occurrences in a list and
- those two occurrences being at specific indices

Also, as in QCheck's list shrinker #268:
> In slogan form the issue can be phrased as: "speed or exhaustiveness - pick one".

Like in that one, we could potentially chose to switch to being more exhaustive below a certain list size.

Issue #182, opened 2021-09-22
0.18 `max_fail` regression


jmid:
Consider this `max_fail` example from the documentation https://c-cube.github.io/qcheck/0.18/qcheck-core/QCheck/index.html
(which says "The 15 smallest counter-example lists will be printed" 😉):

With QCheck 0.17 three counterexamples are printed as expected:
```ocaml
utop # let test = QCheck.(
          Test.make
            ~count:10_000 ~max_fail:3
            (list small_nat)
            (fun l -> l = List.sort compare l));;
val test : Test.t = QCheck.Test.Test <abstr>
utop # QCheck.Test.check_exn test;;
Exception:
test `anon_test_1` failed on ≥ 3 cases:
[1; 0] (after 27 shrink steps)
[1; 0] (after 16 shrink steps)
[1; 0] (after 13 shrink steps)
```

With QCheck 0.18 we only get one counterexample printed unfortunately - a regression:
```ocaml
utop # let test = QCheck.(
          Test.make
            ~count:10_000 ~max_fail:3
            (list small_nat)
            (fun l -> l = List.sort compare l));;
val test : Test.t = QCheck2.Test.Test <abstr>
utop # QCheck.Test.check_exn test;;
Exception:
test `anon_test_1` failed on ≥ 1 cases: [1; 0] (after 13 shrink steps)
```
We should probably add a test for this case... 😉
... or consider using [mdx](https://github.com/realworldocaml/mdx) to ensure that the documentation's examples continue to work as intended.

c-cube:
mdx is cool, but having a unit test is always less flimsy :)

PR #332, opened 2025-03-05
Continue on fail when max_fail is gt 0

Pat-Lafon:
I think I ran into #182, where I want to set `max_fail` to be some value greater than 1.

In my limited sleuthing, I believe the check could have been accidentally removed in https://github.com/c-cube/qcheck/commit/3a9bf2e865665d83c5ed82fc32511f555db2587e#diff-07dbdff6f84a44f565ad4871c5790ad598ceda123f0b8190a711d303d3be28efR1718-R1720 (Hmm, it's highlighted in the link, but it doesn't automatically go to the corresponding line number which is 1767 in file `src/core/QCheck2.ml`). Here, it use to be that if some small parameter and `state.cur_max_fail > 0`, then it would continue otherwise yield. I've just added that latter back in. In the current codebase, it doesn't seem like `cur_max_fail` does anything(4 matches for searching this keyword).

Besides just running this locally and seeing that in my example I can report multiple failures in the failure column, I haven't looked into how to test this. I also haven't confirmed is this actually solves #182 other than seeing that the suspected commit was in April 2021 and the issue was filed in September 2021

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## NEXT RELEASE
 
-- ...
+- Restore `Test.make`'s `max_fail` parameter which was accidentally broken in 0.18
 
 ## 0.24
 
--- a/src/core/QCheck2.ml
+++ b/src/core/QCheck2.ml
@@ -884,7 +884,7 @@ module Gen = struct
     make_primitive
       ~gen:(fun st -> gen st |> Tree.root)
       ~shrink
-  
+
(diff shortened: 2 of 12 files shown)


jmid:
Thanks a bunch for digging into this issue and suggesting a fix :pray: :slightly_smiling_face: 

I've added a couple of tests, filed updated expect test outputs for 32+64 bit builds, and added a CHANGELOG entry.

jmid:
Cool, thanks again :slightly_smiling_face:

PR #334, opened 2025-03-06
CI: Build and test on 5.3.0 too - and target Apple silicon macOS

jmid:
This is the latest and most stable 5.* release after all :slightly_smiling_face:


jmid:
The latest commit excludes macOS running 4.08, as OCaml ARM64 support hasn't been backported that far back
https://github.com/ocaml/ocaml/pull/10026

PR #331, opened 2025-03-04
Prepare for 5.2 AST bump

patricoferris:
In preparation for https://github.com/ocaml/opam-repository/pull/27478 there is WIP documentation about the AST bump on the ppxlib wiki: https://github.com/ocaml-ppx/ppxlib/wiki/Upgrading-to-ppxlib-0.36.0

Issue #276, opened 2023-05-01
Switch to ocaml-ci


jmid:
https://github.com/ocurrent/ocaml-ci is coming along nicely and offers tests on a range of platforms and OCaml versions,
so we might consider switching to it.

Advantages:
- works with the OCaml-ecosystem, meaning less workflow files to maintain (looking at you #275)
- more platform combinations will be exercised, allowing to build confidence in a PR
- non-portable issues will be caught early rather than when I bang my head against the errors from opam-ci at release time

Disadvantages:
 - no Windows testing ATM AFAIK (we would keep a Windows GitHub actions workflow for that though)

PR #335, opened 2025-03-06
Add CI testing on 32-bit i386 of OCaml 4.14 and 5.3 via Docker containers

jmid:
This PR adds CI tests on 32-bit i386 of OCaml 4.14 and 5.3 via Docker containers.

Background:
The PRNG changed in OCaml5, meaning outputs differ between 4 and 5, even when run with the same seed.
On top, OCaml supports 32-bit (on OCaml5 only via bytecode). Again, this means different output behaviour between 32-bit and 64-bit, resulting in 4 combinations overall, that QCheck supports.

Our current CI only tests on 64-bit, thus leaving a blind angle.
More than once have I discovered out-of-date 32-bit expect test files during a release to the opam-repo, which tests 32-bit in its CI. This is a bit late to catch something that ideally would come with each PR.
This PR thus offers exactly that: CI testing of 32-bit builds via Docker containers.
I've only added 2 entries - one representative (4.14 and 5.3) from each side of the PRNG split.

The addition battles a bit to make the ends meet between the prebuilt opam-images using `HOME=/home/opam` and an accompanying `opam` user with `~/.opamrc` etc, and GitHub actions insisting on adding `-e "HOME=/github/home"` last when invoking `docker` :shrug:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,10 +32,39 @@ jobs:
             ocaml-compiler: "4.12"
     runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@v4
-    - uses: ocaml/setup-ocaml@v3
-      with:
-        ocaml-compiler: ${{ matrix.ocaml-compiler }}
-    - run: opam install . --deps-only --with-test
-    - run: opam exec -- dune build
-    - run: opam exec -- dune runtest
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


jmid:
FTR, these can just be removed again if we move to ocaml-ci #276

jmid:
The #331 CI runs were indicating that we were missing an `opam update` in the workflows.

I've added one here - to both 64-bit and 32-bit workflows. 

I'll go ahead and merge, so that #331 can benefit.

PR #344, opened 2025-05-28
Fix QCheck2 float_range operator with negative bounds

rmonat:
Example crash fixed:
> # QCheck_base_runner.run_tests [QCheck2.Test.make ~count:100000 ~print:QCheck2.Print.float (QCheck2.Gen.float_range (-2.) (-1.)) (fun _ -> false)];;
> random seed: 143173129
> Exception:
> Invalid_argument
>  "Gen.float_bound_inclusive: origin value -1. is lower than low value 0.".

I also think we could generate `float_range` to avoid the range limit currently baked in.

I don't know a good way to generate the other expected files for alternative platforms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## NEXT RELEASE
 
+- Fix QCheck2 `float_range` operator which would fail on negative bounds
 - Fix `QCHECK_MSG_INTERVAL` not being applied to the first in-progress message
 
 ## 0.25
--- a/src/core/QCheck2.ml
+++ b/src/core/QCheck2.ml
@@ -445,6 +445,7 @@ module Gen = struct
                    ~origin:(Option.value ~default:(pick_origin_within_range ~low ~high ~goal:0.) origin)
                    ~low
                    ~high in
+    let origin = origin -. low in
     (float_bound_inclusive ~origin (high -. low))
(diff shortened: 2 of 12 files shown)


jmid (approved):
Thanks for the report and fix! :slightly_smiling_face: 
I appreciate you adding a test to both the QCheck and QCheck2 test suites :pray: 

The ocaml4.64 expect file was reporting 1 shrink step on the QCheck test, causing an expect test failure there, so I've pushed an update. I also pushed ocaml5.64 expect files, just produced under a regular 5.3 switch.

To produce 32-bit expect files, for now I've settled on producing them under docker.
In a qcheck repo, the following will run a 32-bit image of (in this case 4.14) with the `qcheck` directory suitably mounted:
```
docker run -v .:/home/opam/qcheck -it --platform linux/i386 ocaml/opam:debian-12-ocaml-4.14
```
One then needs to install `dune` and `alcotest`, run the test, and copy out expect output files from `_build/default` (and then similarly for OCaml5). I'm sure this can be improved, e.g., by writing a Dockerfile by hand.

Installing `alcotest` in the above however failed, due to `ocamlfind` failing to download https://github.com/ocaml/ocamlfind/issues/108 so that required an additional workaround. I've thus pushed updated 32-bit expect outputs as well, along with a CHANGELOG entry.

Thanks again!

PR #345, opened 2025-06-09
Make `run_tests` output progressive

rmonat:
I was tired of having to wait for the whole tests to finish to get any output. This PR factors computation of results and statistics in a single list traversal, but now the output of results is progressive. 

I find this more convenient but I would understand if there are good reasons not to do it that way :)

### DIFF
--- a/src/runner/QCheck_base_runner.ml
+++ b/src/runner/QCheck_base_runner.ml
@@ -431,7 +431,6 @@ let run_tests
     in
     Res (cell, r)
   in
-  let res = List.map aux_map l in
   let aux_fold (total, fail, error, warns) (Res (cell, r)) =
     let warns = warns + List.length (R.get_warnings r) in
     let acc = match R.get_state r, QCheck2.Test.get_positive cell with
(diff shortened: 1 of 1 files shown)


jmid:
I just thought of another thing: :thinking: 
- When running a list of tests without the `--verbose / -v` option, users still need enough context printed to understand, e.g., an error. So an option is to print those differently depending on running verbose (with progressive output providing context) and without (no context provided)

rmonat:
From my understanding of `qcheck`, this change makes the output progressive, but I'm not sure it concerns all outputs. I wasn't aware of all these kind of ouputs. I agree that some options could help too!

> With failing tests in the mix, the header generated  error   fail   pass /  total     time test name flows away from the lines below, intended to be nicely aligned under it. Should we go away from printing a header? Or reprint it after inferring output?

I'm not sure it's easy to do properly without new dependencies and in a cross-platform fashion, but we could probably keep some space for the header and print the rest/logs afterwards?


> When there is at most one Failure, Error, Info or Messages(!?) for each test having just run, with progressive output we should be able to remove some redundancy (Failure ... Test ... failed, Messages ... Messages for ... and repeatedly printing the same test name). The situation is just different from when they are printed apart from each other and hence requires repetition ("Here is the counterexample for test X42 which we ran above")

Agreed!

> Do we want labeled and numeric stats printed progressively as well, or should they continue to go at the bottom?
(I personally think either is fine, but would like the choice to be a conscious decision)

Is this displayed by default? It could be an option? Or displayed after the summary, and the logs/messages?

PR #357, opened 2025-10-10
Add QCheck.Shrink.float

### LINKED ISSUE #89: Add a float shrinker
The `Shrink` module is missing a `float` shrinker.
Part of the issue is figuring out what the shrinker should do, perhaps by researching the heuristics other frameworks use. For example, a halving heuristic is not going to work well for floats.

jmid:
This adds a `float` shrinker for `QCheck`. Fixes #89.

After having experimented, struggled, and abandoned a different binary design based on `Float.frexp`, this shrinks
- towards `1.` or `-1.`,
- based on the printed, decimal "to-string" rendering of a `float` in [scientific notation](https://en.wikipedia.org/wiki/Scientific_notation) `7.29806027239e+186`,
- reducing the significand `7.29806027239` and exponent `186` separately, and
- preferring fewer decimals, and a shorter string rendering (with fewer chars)

For a `float`, "smaller" does not mean "simpler". I would argue that `1.` is a simpler counterexample than `0.5`, `0.25`, `0.125`, `0.0625`, ... as one would find by repeated halving and moving closer and closer to `0.` 
FTR, 'shrink-towards-0.' is the strategy of the current `QCheck2.Gen.float` generator's shrinker
https://c-cube.github.io/qcheck/0.26/qcheck-core/QCheck2/Gen/index.html#val-float

While it initially seemed clunky to me to have to go though `string`s, this is nevertheless how counterexamples are presented to users, so in retrospect I believe it makes sense. After all, using fewer 1 bits in binary does not necessarily result in a shorter printed rendering.  

The implementation is backed by a range of unit tests tracking individual shrink steps.

The updates to the expect test outputs also reveal a nice quality-of-life improvement to end-users, without spending too many shrink steps to arrive at the final counterexamples.

Finally, I've tried to be generous with comments to make the code and the heuristic accessible to others.

Unless there are objections or input causing drastic changes, I plan to port this heuristic to `QCheck2` (in a separate PR).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## NEXT RELEASE
 
-- ...
+- Add `QCheck.Shrink.float` and enable shrinking for `QCheck.float`
+
 
 ## 0.26
 


jmid:
All green still, so I'll go ahead and merge.

PR #358, opened 2025-10-24
Small fixes for OxCaml

vouillon:
Somehow, OxCaml does not like these partial applications:
```
File "src/core/QCheck.ml", line 1266, characters 45-55:
1266 |   let small = _opt_map_or ~d:Array.length ~f:array_sum_ a.small in
                                                    ^^^^^^^^^^
Error: This expression has type ('a -> int) -> 'a array -> int
       but an expression was expected of type
         ('a -> int) -> 'b array @ immutable -> int
       Type 'a array -> int is not compatible with type
         'b array @ immutable -> int
```


c-cube:
Thank you, that's very reasonable :)

Issue #103, opened 2021-03-11
Don't raise an error on test failure


Sir4ur0n:
When using QCheck with Alcotest (`qcheck-alcotest`), when an error occurs, the logs are rather unceremonious, and in particular, the logs show a confusing (and arguably useless) error, see e.g.:
```
test `My awesome QCheck test` failed on ≥ 1 cases:
My-Pretty-Printed-Value

[exception] test `My awesome QCheck test` failed on ≥ 1 cases:
My-Pretty-Printed-Value

            Raised at file "src/core/QCheck.ml", line 1726, characters 6-38
            Called from file "src/alcotest-engine/core.ml", line 277, characters 17-23
            Called from file "src/alcotest-engine/monad.ml", line 34, characters 31-35
```

I think there are several things that can be improved:
1. The useful part of the message is duplicated (test name, pretty printed value, number of cases, etc.)
2. The error (`Raised at...`) is unnecessary (and actually confusing to users, at least it was to me :smile: )
3. Some content around the pretty print of generated values (e.g. if the generated value is the empty string `""`, I think it is easy to not understand what is going on)

Can the situation be improved? I have no idea if this is a QCheck problem, a QCheck + Alcotest problem, or an Alcotest problem (e.g. limited Alcotest integration API).

Disclaimer: I am coming from the Haskell ecosystem where the [Hedgehog](https://hackage.haskell.org/package/hedgehog) library has excellent error reporting, see e.g. [this blog article](https://jacobstanley.io/5-tips-for-better-hedgehog-tests/):
```
   ┏━━ test/Article.hs ━━━
44 ┃ prop_roundtrip :: Property
45 ┃ prop_roundtrip =
46 ┃   property $ do
47 ┃     x <- forAll genUser
   ┃     │ User { userId = "00000" , userName = "stephanie" }
48 ┃     tripping x Aeson.encode Aeson.eitherDecode
   ┃     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ┃     │ ━━━ Intermediate ━━━
   ┃     │ "{\"userName\":\"stephanie\",\"userId\":\"00007\"}"
   ┃     │ ━━━ - Original) (+ Roundtrip ━━━
   ┃     │ - Right User { userId = "00000" , userName = "stephanie" }
   ┃     │ + Right User { userId = "00007" , userName = "stephanie" }
```

[...]

PR #405, opened 2026-02-10
Remove duplicate printing from Alcotest backend

ptrichr:
Hi, I noticed that when using the `Alcotest` backend for `QCheck`, there is duplicated output pretty printer output on test failures - one set from `Alcotest` and one set from `QCheck`. 

I believe a similar issue is mentioned in #103. I think I have corrected this issue in this PR, however, I would love a maintainer to sanity check me, as I'm not incredibly familiar with the `QCheck` codebase.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@
   - `Observable.map`
   - `TestResult.stats` `TestResult.warnings` `TestResult.collect`
 - Clarify documentation for the `int*` and `nat*` generators
+- Disabled duplicated pretty-printed feedback when using `QCheck_alcotest` runner
 
 
 ## 0.91 (2025-12-21)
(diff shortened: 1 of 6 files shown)


jmid (commented):
Thanks, this makes sense to me 👍 

Triggering a CI run I see this failing expect tests, because "double printing" is missing - effectively giving the example test I was going to ask about.

Can you update and commit the updated expect tests so that `dune runtest example/alcotest` passes locally?

Also, a CHANGELOG entry would be welcome!

ptrichr:
Removed the duplicated output from expected. Additionally, locally, it's also complaining about the size of the boxes (though it didn't do that in the previous CI run), so I'm going to see if that changed as well.

jmid (approved):
LGTM, thanks for the contribution! 🙏

PR #406, opened 2026-02-11
Document the ppx `@weight` attribute

giltho:
I originally forked the repo to create a `[@ignore]` attribute to the PPX, and realised that the more general `[@weight 0]` was already implemented.

PS: qcheck is really awesome!

### DIFF
--- a/src/ppx_deriving_qcheck/README.md
+++ b/src/ppx_deriving_qcheck/README.md
@@ -242,6 +242,28 @@ let gen_color =
      (1, (QCheck.Gen.pure `Green))] : color QCheck.Gen.t)
 ```
 
+### Overwrite weights
+
+If you wan't to specify your own weigth for one of the variants, you can
+add an attribute to the type.
+
+For instance, the following will never generate any `Blue` variant:
+```ocaml
+type color =
+  | Red
+  | Blue [@weight 0]
+  | Green
+[@@deriving qcheck]
+
+(* ==> *)
+
+let gen_color =
+  QCheck.Gen.oneof_weighted
+    [(1, (QCheck.Gen.pure Red));
+     (0, (QCheck.Gen.pure Blue));
+     (1, (QCheck.Gen.pure Green))]
+```
+
 ## Recursive variants
 * Recursive variants
 ```ocaml
--- a/src/ppx_deriving_qcheck/README.md
+++ b/src/ppx_deriving_qcheck/README.md
@@ -244,7 +244,7 @@ let gen_color =
 
 ### Overwrite weights
 
-If you wan't to specify your own weigth for one of the variants, you can
+If you wan't to specify your own weight for one of the variants, you can
 add an attribute to the type.
 
 For instance, the following will never generate any `Blue` variant:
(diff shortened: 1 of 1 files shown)


src/ppx_deriving_qcheck/README.md
```diff
@@ -242,6 +242,28 @@ let gen_color =
      (1, (QCheck.Gen.pure `Green))] : color QCheck.Gen.t)
 ```
 
+### Overwrite weights
+
+If you wan't to specify your own weight for one of the variants, you can
```
jmid:
```suggestion
If you want to specify your own weight for one of the variants, you can
```

src/ppx_deriving_qcheck/README.md
```diff
@@ -242,6 +242,28 @@ let gen_color =
...
+  | Green
+[@@deriving qcheck]
+
+(* ==> *)
+
+let gen_color =
+  QCheck.Gen.oneof_weighted
+    [(1, (QCheck.Gen.pure Red));
+     (0, (QCheck.Gen.pure Blue));
+     (1, (QCheck.Gen.pure Green))]
```
jmid:
```suggestion
For instance, the following will generate a `Blue` variant twice as often as a `Red` or `Green` variant:
```ocaml
type color =
  | Red
  | Blue [@weight 2]
  | Green
[@@deriving qcheck]

(* ==> *)

let gen_color =
  QCheck.Gen.oneof_weighted
    [(1, (QCheck.Gen.pure Red));
     (2, (QCheck.Gen.pure Blue));
     (1, (QCheck.Gen.pure Green))]
```


Using weights to disable a variant is clever - and should perhaps be mentioned.
I think it would make sense to illustrate the more common use case of skewing a distribution, hence the above suggestion.

jmid (commented):
Thanks a bunch - this should indeed be documented!

I've made a few minor suggestions.

src/ppx_deriving_qcheck/README.md
giltho: Makes sense yeah. I had the 0 example in mind because this is what I was looking for in the first place but it's true that it's not the main use case 😅

giltho:
Sorry for the typo as well, bit sloppy not sure why I didn't catch it 😅

src/ppx_deriving_qcheck/README.md
```diff
@@ -242,6 +242,28 @@ let gen_color =
...
+
+(* ==> *)
+
+let gen_color =
+  QCheck.Gen.oneof_weighted
+    [(1, (QCheck.Gen.pure Red));
+     (2, (QCheck.Gen.pure Blue));
+     (1, (QCheck.Gen.pure Green))]
+```
+
```
jmid, line 266:
```suggestion

The `weight` attribute can also be handy to disable generation of a variant by annotating it with `[@weight 0]`.

```
I think your original example with `@weight 0` is neat and should be mentioned too, as it may come in handy for other users.

src/ppx_deriving_qcheck/README.md
jmid: Weight 0 is a good use case indeed!

In the latest commit I added a mention of it. With both use cases documented this is in good shape!

jmid:
Thanks again!

PR #410, opened 2026-05-01
fix the overflow at (https://github.com/c-cube/qcheck/issues/408)

### LINKED ISSUE #408: `Test.make_cell` `~count` overflow bug
I think I hit an overflow bug when the `count` is too high, likely culpript (https://github.com/c-cube/qcheck/blob/7a70422c43600879c1edf11f17e943a9543b14a4/src/core/QCheck2.ml#L1621 or https://github.com/c-cube/qcheck/blob/7a70422c43600879c1edf11f17e943a9543b14a4/src/core/QCheck2.ml#L1648).

Context: I maintain [ETNA](http://github.com/alpaylan/etna-cli) which also has qcheck workloads, I hit the error when I was working on a migration for one of them.

```ocaml
(* QCheck2.Test.check_cell silently runs zero iterations when
   `count` exceeds (max_int - 199).

   On a 64-bit system, max_int = 2^62-1 = 4611686018427387903.
   The threshold is empirically max_int - 199 = 4611686018427387704
   — at that count and above, `get_count_gen result` returns 0 even
   though the property is `fun _ -> false` and would fail at the first
   iteration if the loop ran.

   Bisect with this binary:
     dune exec ./repro.exe -- <count>

   Observed (qcheck-core 0.26 on ocaml 5.3.0 / 4.14.0):
     count=1                     -> tested=1
     count=10000000              -> tested=1
     count=4611686018427387703   -> tested=1     (max_int - 200)
     count=4611686018427387704   -> tested=0     (max_int - 199, BUG)
     count=4611686018427387903   -> tested=0     (max_int)

   This strongly suggests a `count + K` arithmetic step inside
   check_cell (or its setup) overflows to a negative int for
   `count > max_int - K`, after which the iteration loop exits
   immediately. K appears to be 200 -- likely a corner-cases buffer
   added to the user-supplied count.
*)
open QCheck2

let () =
  let count =
    if Array.length Sys.argv < 2 then max_int
    else int_of_string Sys.argv.(1)
  in
  let cell =
    Test.make_cell ~name:"never_true" ~count Gen.nat (fun _ -> false)
  in
  let result = Test.check_cell cell ~rand:(Random.State.make [| 0 |]) in
  Printf.printf
    "count=%d tested=%d\n"
    count (TestResult.get_count_gen result)
```

alpaylan:
I applied the suggested fix, added a regression test, ran and checked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@
   - `TestResult.stats` `TestResult.warnings` `TestResult.collect`
 - Clarify documentation for the `int*` and `nat*` generators
 - Disabled duplicated pretty-printed feedback when using `QCheck_alcotest` runner
+- Fixed the overflow bug when `~count > max_int - 200` affecting `Test.{make_cell,make,make_neg}` in both QCheck and QCheck2
 
 
 ## 0.91 (2025-12-21)
--- a/src/core/QCheck2.ml
+++ b/src/core/QCheck2.ml
@@ -1618,7 +1618,7 @@ module Test = struct
     let count = global_count count in
     let long_factor = global_long_factor long_factor in
     let positive = not negative in
-    let max_gen = match max_gen with None -> count + 200 | Some x->x in
+    let max_gen = match max_gen with None -> max count (count + 200) | Some x->x in
     {
       law;
       gen;
@@ -1645,7 +1645,7 @@ module Test = struct
     let positive = not negative in
     (* Make a "fake" QCheck2 arbitrary with no shrinking *)
     let fake_gen = Gen.make_primitive ~gen ~shrink:(fun _ -> Seq.empty) in
-    let max_gen = match max_gen with None -> count + 200 | Some x->x in
+    let max_gen = match max_gen with None -> max count (count + 200) | Some x->x in
     {
       law;
       gen = fake_gen;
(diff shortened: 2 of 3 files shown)


jmid (approved):
This is excellent - thanks a bunch - also for adding a nice regression test! 🙏😃 

Can you add a CHANGELOG entry, mentioning that the fix concerns an issue with both `Test.make_{test,cell,...}` - and in both `QCheck` and `QCheck2`?

alpaylan:
Done, thank you for the kind comments!

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@
   - `TestResult.stats` `TestResult.warnings` `TestResult.collect`
 - Clarify documentation for the `int*` and `nat*` generators
 - Disabled duplicated pretty-printed feedback when using `QCheck_alcotest` runner
+- Fixed the overflow bug when `~count > max_int - 200` that affected both `Test.make_{test,cell,...}` in both QCheck and QCheck2
```
jmid:
```suggestion
- Fixed the overflow bug when `~count > max_int - 200` affecting `Test.{make_cell,make,make_neg}` in both QCheck and QCheck2
```

jmid: It is probably a good idea to write out the affected functions.

Also: there was a CI issue - can you rebase on top of #412?

alpaylan: I believe it should be fine now

jmid:
Thanks!

PR #411, opened 2026-05-06
Add doc warning about ==> precedence

### LINKED ISSUE #407: Precedence of QCheck2.(==>)
Apparently, `QCheck2.(false && true ==> true)` is `false`, whereas in propositional calculus, it is standard that conjunction have precedence over implication. This is a bit puzzling and should be at least documented in the documentation of `==>` and ideally fixed.

The manual https://ocaml.org/manual/5.4/expr.html#ss:precedence-and-associativity says that indeed an infix operator `=...` is more tight than `&...`, which is consistent with the above observation but tends to imply that there is no easy fix without changing the symbol `==>`.

Note: the documentation https://ocaml.org/docs/operators#operator-associativity-and-precedence looks incorrect since it puts both operators in the same precedence group, which is furthermore said to be left-associative: this contradicts the above observation.

jmid:
This PR adds documentation warnings about how the precedence of `==>` interacts badly with the logical connectives `&&` and `||`.

Fixes #407

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 - Clarify documentation for the `int*` and `nat*` generators
 - Disabled duplicated pretty-printed feedback when using `QCheck_alcotest` runner
 - Fixed the overflow bug when `~count > max_int - 200` affecting `Test.{make_cell,make,make_neg}` in both QCheck and QCheck2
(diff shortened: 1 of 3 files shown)


jmid:
(rebased after having merged #412)

Issue #413, opened 2026-05-08
Generating dependent pairs


benjub:
I have had several times to generate dependent pairs in my property checks: for instance, pairs formed of an array and a valid index.

To do that, I used the function
```
let dep_pair a f = a >>= fun x -> tup2 (pure x) (f x)
```
(where `QGheck2.Gen` is open), rather than first generating all pairs and then discarding the unwanted ones using `assume`.

The above example can then be obtained with:
```
let array_index =
  dep_pair
    (array_size (int_range 1 100) nat_small)
    (fun t -> int_bound (pred (Array.length t)))
```
My use cases included more complex pairs of structures bound by relations, but I think the above example shows the basic idea.

A function like `dep_pair` may be worth adding to `QCheck2.Gen`.

PR #414, opened 2026-05-12
Add Gen.dep_pair combinator for dependent pair generation

kimjune01:
## Summary

- Adds `Gen.dep_pair : 'a t -> ('a -> 'b t) -> ('a * 'b) t` to QCheck2, which generates pairs where the second element depends on the first. This avoids the need to use `assume` to filter invalid combinations (e.g. generating an array and a valid index into it).
- Includes documentation with usage example and a test that verifies generated array-index pairs are always valid.

## Test plan

- [ ] `dune build` succeeds
- [ ] `dune runtest` passes, including the new `dep_pair` test in `QCheck2_unit_tests.ml`
- [ ] Verify the example in the `.mli` documentation compiles

### DIFF
--- a/src/core/QCheck2.ml
+++ b/src/core/QCheck2.ml
@@ -733,6 +733,9 @@ module Gen = struct
 
   let pair (g1 : 'a t) (g2 : 'b t) : ('a * 'b) t = liftA2 (fun a b -> (a, b)) g1 g2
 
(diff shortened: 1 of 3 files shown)


jmid:
This is apparently part of a hit list experiment to target LLMs at various open source repos:
https://june.kim/speedrunning-open-source
Sorry, I don't want to be part of that, nor spend my spare time reviewing your slop where no effort has been invested beyond a copy-paste from #413. User blocked.

